PR #155, opened 2019-01-18
Implement getSharedLibs() using "debug/elf"

mkhon:
There is also a support for NANOS_TARGET_ROOT env variable (if set - points to the Linux chroot environment) - it is currently commented out.

There's corresponding code in mkfs (also ifdef'ed out under DEV) which also implements NANOS_TARGET_ROOT.

If NANOS_TARGET_ROOT is compiled-in ops/mkfs can be used on macOS to build images for dynamically linked Linux programs (given that chroot environment is provided).

chroot environment can be pre-created on Linux with e.g. "debootstrap"


eyberg:
does that env var force a chroot on osx or just optional?

mkhon:
@eyberg it is optional (if not set ops/mkfs behave as before) and it is even not compiled-in by default (although I would prefer to have it compiled-in by default)

lepton/elf.go
```diff
@@ -0,0 +1,140 @@
...
+			break;
+		}
+	}
+
+	// fmt.Printf("%s -> %s\n", path, targetPath)
+	return targetPath, nil
+}
+
+func findLib(libDirs []string, path string) (string, error) {
+	if path[0] == '/' {
```
tijoytom: When is path[0] == '/'? Is it when on Mac?

mkhon: @tijoytom I've just tried to follow ld.so(8) man page:

```
If a shared object dependency does not contain a slash, then it is searched for in the following order:
...
```

Issue #167, opened 2019-01-23
 temporary files cause `ops run` to fail with permission errors in multi-user systems.


ExternalReality:
When `ops run` executes it makes 2 temp files; `/tmp/hostname` and `/tmp/resolv`. These files are created with the permissions mode 644 and are owned by the user who ran the command. If another user on the same system should run `ops run` then the command will fail with permission errors (on those files) unless the temp files are removed or the permissions are corrected. What is the appropriate solution?

eyberg:
not even confident these belong here but for now this could prob refactored to use https://golang.org/pkg/io/ioutil/#TempFile instead - @tijoytom thoughts?

the other thing is that the manifest is only a temporary file that's used in flight so some cleanup could be done as well

eyberg:
https://github.com/nanovms/ops/issues/196

PR #170, opened 2019-01-24
change perm for temp file; give rw for everyone

tijoytom:


tijoytom:
https://github.com/nanovms/ops/issues/167

Issue #173, opened 2019-01-25
Make all qemu parameters configurable, there some hard-coded


tijoytom:
Better if the config can be generic ie not tied to a specific hyper visor

ExternalReality:
How about  package dependencies using [gopkg.in](https://golang.github.io/dep/) and pulling in intels [qemu package](https://github.com/intel/govmml)? Its small its simple and its got all the types laid out quite nicely.

tijoytom:
I am talking about the command line parameter passed to qemu not package dependency.  This link  qemu package is broken, so not sure what it means.

tijoytom:
I see, i don't see any particular reason to use intel package, it just supports qemu and probably not maintained too. It's pretty straight forward to create our own wrapper, we already have the interfaces for it( ie the Hypervisor interface), just need to wire them up. And about deps, let's not introduce something unless we absolutely need it.Keep it simple as much as we can.

ExternalReality:
https://github.com/nanovms/ops/pull/182/files

PR #176, opened 2019-01-26
Refactor Qemu code for easy tranformation from run configuration.

ExternalReality:
This is a refactor. The idea is to make it easy to simply transform a `RunConfig` into a "Qemu" struct (or just to take a `RunConfig` and convert it into qemu command line string). The next step is to get the correct parameters onto the cli and into the `RunConfig` to complete issues https://github.com/nanovms/ops/issues/173 and https://github.com/nanovms/ops/issues/177.

I have unit test and I have successfully run the commands with the various options but if you agree with this commit please eye the rendering functions to ensure I did not miss anything that will break the rendering of a qemu command.

### DIFF
--- a/lepton/qemu.go
+++ b/lepton/qemu.go
@@ -7,30 +7,107 @@ import (
 	"strings"
 )
 
+const qemuBaseCommand = "qemu-system-x86_64"
+
 type drive struct {
 	path   string
 	format string
+	iftype string
+	index  string
+}
+
+type device struct {
+	driver   string
+	mac      string
+	netdevid string
 }
 
 type netdev struct {
-	nettype string
-	id      string
-	ifname  string
-	mac     string
+	nettype    string
+	id         string
+	ifname     string
+	script     string
+	downscript string
+	hports     []portfwd
+}
+
+type portfwd struct {
+	port  int
+	proto string
+}
+
+type display struct {
+	disptype string
+}
+
+type serial struct {
+	serialtype string
 }
 
 type qemu struct {
-	cmd    *exec.Cmd
-	drives []drive
-	ifaces []netdev
+	cmd     *exec.Cmd
+	drives  []drive
+	devices []device
+	ifaces  []netdev
+	display display
+	serial  serial
 }
 
-func (q *qemu) addDrive(path string, format string) {
-
+func (d display) String() string {
+	return fmt.Sprintf("-display %s", d.disptype)
 }
 
-func (q *qemu) addNetworkDevice(n netdev) {
+func (s serial) String() string {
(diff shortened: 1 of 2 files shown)


lepton/qemu.go
```diff
@@ -7,37 +7,112 @@ import (
...
+func (nd *netdev) render() []string {
+	var str string
+	str = fmt.Sprintf("%s,id=%s", nd.nettype, nd.id)
+	if len(nd.ifname) > 0 {
+		str += "," + fmt.Sprintf("ifname=%s", nd.ifname)
+	}
+	if len(nd.script) > 0 {
+		str += "," + fmt.Sprintf("script=%s", nd.script)
+	}
+	if len(nd.downscript) > 0 {
```
tijoytom: If there is no downscript you need to explicitly set it to 'no' to avoid qemu trying to execute script from a default location and then asking to be root.

lepton/qemu.go
```diff
@@ -7,37 +7,112 @@ import (
 	"strings"
 )
 
+const qemuBaseCommand = "qemu-system-x86_64"
+
+type QemuRender interface {
```
tijoytom: You don't need a new interface, implement String() method for all types.

lepton/qemu.go
```diff
@@ -66,32 +141,30 @@ func (q *qemu) Start(rconfig *RunConfig) error {
...
 func (q *qemu) Args(rconfig *RunConfig) []string {
-	// TODO : this should come from q.drives and ifaces
+	// todo create a qemu from serial
 	args := []string{}
-
-	boot := []string{"-drive", "file=image,format=raw,index=0"}
-	storage := []string{"-drive", "file=image,format=raw,if=virtio"}
-	var net []string
-	if rconfig.Bridged {
-		net = []string{"-device", "virtio-net,mac=7e:b8:7e:87:4a:ea,netdev=n0",
```
ExternalReality: @tijoytom did you mean for this to have (down)script=no?

ExternalReality:
There is another modification that has to be done to this.

lepton/qemu.go
ExternalReality: OK, I see what happened here. For the 'user' device type the script parameters are invalid. I added that logic to the `netdev` String() method and added a unit test. PR should be good now. populating a `qemu` struct with info from a `RunConfig` should be enough to render a good command string.  `-nodefaults`, `-no-reboot`, `-device isa-debug-exit`, I'm guessing these things are not variable and will be used in all cases, so they are left hard-coded.

tijoytom: you can leave them hard-coded.

PR #182, opened 2019-01-31
Make params configurable

ExternalReality:
Here params are configurable modulo knowing what the extra flag and config file attributes should look like.

### DIFF
--- a/lepton/qemu.go
+++ b/lepton/qemu.go
@@ -1,6 +1,7 @@
 package lepton
 
 import (
+	"crypto/rand"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


lepton/qemu.go
```diff
@@ -143,36 +145,111 @@ func (q *qemu) Start(rconfig *RunConfig) error {
...
+		driver:   "virtio-net",
+		netdevid: id,
+	}
+	ndv := netdev{
+		nettype: devType,
+		id:      id,
+	}
+	if devType != "user" {
+		if mac == "" {
+			dv.mac = generateMac()
```
ExternalReality: Generate a uinicast mac address in lieu of a specified mac.

PR #185, opened 2019-01-31
simple package list filtering.

ExternalReality:
Really simple filtering with a regex.

Let me get a test in there.

### DIFF
--- a/ops.go
+++ b/ops.go
@@ -5,17 +5,19 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"os/exec"
 	"path"
+	"regexp"
 	"runtime"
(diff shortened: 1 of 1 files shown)


ops.go
```diff
@@ -367,15 +373,35 @@ func cmdListPackages(cmd *cobra.Command, args []string) {
...
 	table.SetRowLine(true)
 
-	for key, value := range pacakges {
+	var r *regexp.Regexp
+	var filter bool
+	if len(searchRegex) > 0 {
+		filter = true
+		r, err = regexp.Compile(searchRegex)
+		if err != nil {
+			panic(err)
```
tijoytom: Don't need to panic here, if regex is not valid continue as if there is no filter.

Issue #154, opened 2019-01-17
include update command to check for ops {boot,mkfs,staging3} upates


eyberg:
~~* mv {boot,mkfs,staging3} to https://github.com/nanovms/ops/issues/153~~

* add sha2  for these and throw into some sort of manifest

* update can consult latest manifest and if out of sync download

anikitenko:
Regarding MD5, I have some ideas:
- Without having separate manifest for system images: problem here is that there is always need to download images and then compare MD5 sums.. so.. sounds not so good
- Having separate manifest (for example system.json) - avoids downloading files each time but.. having 2 manifests, as I can see, will require to always make sure that manifests are up to date (according to  logic that I see)
Probably "update" function should be extended to be able to update manifests too (as @eyberg already stated).
Please tell me if I'm on correct way of thinking.. :)

eyberg:
right now there are 2 release channels - 'nightly' (which is whatever is found in master, eg: bleeding edge) and releases we cut manually which is what general population gets, as of right now we aren't keeping more than one of each - maybe in the future we will but not right now

I think right now it's prob. useful to zip these 3 things up into a tarball and md5 that and also host the md5 inside the bucket so that ppl can be assured their stuff wasn't tampered with

the original request was just to move the core stuff into a common area && implement light-weight versioning so it's not downloading the entire thing each time for a new project

anikitenko:
Okay then, my PR covers first part "move the core stuff into a common area", will think about "implement light-weight versioning"..)

eyberg:
https://github.com/nanovms/ops/issues/307

eyberg:
we've had a update/version check for a while now - closing

Issue #153, opened 2019-01-17
mv {mkfs,staging3,boot} into ~/.ops to cache && hash


eyberg:

eyberg:
@tijoytom has implemented this in latest

PR #195, opened 2019-02-12
Work on issue #154

anikitenko:
Started my work on issue #154. Currently only "mv {boot,mkfs,staging3} to #153" is done. I'll continue working.. ready for comments..)

### DIFF
--- a/lepton/const.go
+++ b/lepton/const.go
@@ -53,7 +53,7 @@ func GetPackageCache() string {
 		if err != nil {
 			panic(err)
 		}
-		PackagesCache = path.Join(home, ".ops/packages")
+		PackagesCache = path.Join(home, OpsDir, "packages")
 		if _, err := os.Stat(PackagesCache); os.IsNotExist(err) {
 			if err := os.MkdirAll(PackagesCache, 0755); err != nil {
 				panic(err)
--- a/lepton/const.go
+++ b/lepton/const.go
@@ -21,11 +21,14 @@ const manifest string = `(
 const Version = "0.2"
 const OpsReleaseUrl = "https://storage.googleapis.com/cli/%v/ops"
 
+// .ops directory
+const OpsDir string = ".ops"
+
 // boot loader
-const BootImg string = ".staging/boot.img"
+const BootImg string = "boot.img"
 
 // kernel
-const KernelImg string = ".staging/stage3.img"
+const KernelImg string = "stage3.img"
 
 // kernel + ELF image
 const mergedImg string = ".staging/tempimage"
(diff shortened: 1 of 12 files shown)


lepton/const.go
```diff
@@ -21,19 +21,22 @@ const manifest string = `(
 const Version = "0.2"
 const OpsReleaseUrl = "https://storage.googleapis.com/cli/%v/ops"
 
+// .ops directory
+const OpsDir string = ".ops"
```
tijoytom: This should be use $HOME/.ops

lepton/image.go
```diff
@@ -249,62 +256,82 @@ func (bc dummy) Write(p []byte) (int, error) {
 	return len(p), nil
 }
 
-func DownloadBootImages() error {
```
tijoytom: Dont delete this function, it's used else where.

lepton/image.go
```diff
@@ -249,62 +256,82 @@ func (bc dummy) Write(p []byte) (int, error) {
...
 
 	fmt.Println("Downloading..", filepath)
 	out, err := os.Create(filepath + ".tmp")
 	if err != nil {
 		return err
 	}
-	defer out.Close()
-	
+	defer func() {
+		if err := out.Close(); err != nil {
```
tijoytom: Probably no need to panic here, Close returns any errors from the underlying listener. I am fine with just out.Close and not checking for error here.

lepton/const.go
anikitenko: Extended a little use of constant OpsDir. It's needed to determine "base" directory in HOME of user. For example, path to `packages` directory is `path.Join(home, OpsDir, "packages")`

lepton/image.go
anikitenko: I didn't delete this function. I just extended it to make sure that `OpsDir` exists and it's name is the same `DownloadBootImages`. I executed tests and checked for use of that function - everything is alright

lepton/image.go
anikitenko: Agree. I removed panics in defer functions but left as following to silence editor warnings like "Unhandled error..":
`defer func() {
		_ = out.Close()
	}()`
and
`defer func() {
		_ = resp.Body.Close()
	}()`

eyberg: yeh i mean those warnings are there for a reason - i'd prefer we handle all errors correctly - it's a go idiom (and just good programming as well)

anikitenko: and you're right too.. there are no panics in that func but are returns.. so should I update that to `return` if `Close` returns error?

tijoytom: Editors are just dump, they don't have any contextual information. So I won't worry too much about some editor plugin warning about it. But anyways just return the error and yes never panic inside the exported function. Let the caller handle it.

anikitenko: Got it, updating PR..)

lepton/image.go
tijoytom: You renamed it right? What I was telling is that it's used outside of this repository. So you won't see any test failure, remember this is a library shared across projects

anikitenko: Got it, returned back function with it's parameters

lepton/image.go
anikitenko: I updated PR with reworked defer funcs

Issue #193, opened 2019-02-11
ops load leaves artifacts from old version of the package


skupr:
When I was creating new ops package, and testing it via 'ops load', I've encountered an issue that despite package is unpacked into .staging directory on each 'load', old package files are not deleted - it only adds new and updates existing files. In the case new package needs some old files deleted (in my case, I've had some files with space, which break mkfs), they still present in .staging/<package_name> directory, which may cause some confusion and incorrect work of the new version of the package.

eyberg:
we have an aux tkt to md5 each pkg -> https://github.com/nanovms/ops/issues/169 - i think this is pretty important for a number of reasons

also - package should be downloaded to ~/.ops/packages - then base images can be created in project root ~/.staging but it makes no sense to download each time to ~/.staging

tijoytom:
We don't download to .staging every-time, we once download to ./ops/packages and then extract to .staging every-time. You can add code to  delete the folder before extracting and it should be good. Reason we extract each time is to make sure that packages are clean. But we can avoid doing the extraction if required.

tijoytom:
https://github.com/nanovms/ops/issues/193

skupr:
Fixed by #200

PR #200, opened 2019-02-13
remove the package directory if it exists

tijoytom:


lepton/package.go
```diff
@@ -26,6 +26,10 @@ func DownloadPackage(name string) string {
 }
 
 func ExtractPackage(archive string, dest string) {
+
+	// Remove the directory if it exits.
+	os.RemoveAll(dest)
```
tijoytom: error is ignored since we don't have to fail if we can't remove the folder.

skupr: This change removes not only package directory, but the whole .staging directory.
This means that boot.img, stage3.img and mkfs will be redownloaded each time.
In the case described at #193, it's not good, as I want to use fresh images - and this change leaves me no other choice than use downloaded images.

Issue #125, opened 2019-01-09
ops gets 


dpreilan:
I am seeing the same error using ops that nanos saw and fixed with:
https://github.com/nanovms/nanos/issues/31

qemu-system-x86_64 --version
QEMU emulator version 2.11.1(Debian 1:2.11+dfsg-1ubuntu7.9)

cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.1 LTS"

mkhon:
@cbarraford documentation fix is not correct: the problem exists in qemu 2.11 which is >= 2.5

mkhon:
-drive file=image,format=raw,if=none,id=drive-virtio-disk0 -device virtio-blk-pci,drive=drive-virtio-disk0

eyberg:
@mkhon what do you think of the suggestion that we should just grab the version of qemu in ops and use one setting vs the other? that to me sounds like the better option cause ubuntu 16.04 installs 2.5 - i think brew installs 3.0 - espc. since this is different from version to version

mkhon:
@eyberg this was initially my thinking (to differentiate on qemu version).

However I found out how KVM runs virtio with one drive (see command line above) - I think we should use this one - it works on all qemu versions.

@skupr will implement this

mkhon:
nanos boot2 will use BIOS after nanovms/nanos#480 is merged

eyberg:
once this is done (since it's in master nanos) we should prob. bump ops version && cut a release (or can wait until gce integration)

PR #201, opened 2019-02-13
Removed ide drive mapping as it's not neeed after nanovms/nanos#480

skupr:
For #125
After change nanovms/nanos#480, there is no need to map the same image twice to qemu. Removed "ide" mapping, as it's not used anymore bo boot.

### DIFF
--- a/lepton/qemu.go
+++ b/lepton/qemu.go
@@ -214,6 +214,7 @@ func (q *qemu) addOption(flag, value string) {
 }
 
 func (q *qemu) setConfig(rconfig *RunConfig) {
+	// add virtio drive
 	q.addDrive(rconfig.Imagename, "virtio")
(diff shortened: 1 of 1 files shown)


eyberg:
we should prob. cut a release of nanos && ops for this change

skupr:
I'll investigate failing test

skupr:
Test is failed because the changes in this patch require new image with nanovms/nanos#480, but the test downloads old images. @eyberg how should we resolve this? Can we upload fresh images to the cloud so they will be picked up by tests?

eyberg:
yeh - @tijoytom - let's see if we can cut a manual release tmo - if that goes well we can merge this and then cut a new release of ops itself

tijoytom:
@skupr Everything looks good now.

PR #206, opened 2019-02-13
Add binary/source installation instructions.

ExternalReality:
Add `Installation` section with `source` and `binary` sub-sections.


eyberg:
cool - one thing is that we are starting to stuff meta-data and such inside of ~/.ops which the clone from git doesn't setup or deal with - this might be a combination of additional makefile or clarification on readme - not sure

ExternalReality:
Forgive me but I don't understand. Isn't `~/.ops` created and populated on first run. Would that not be more robust than creation and population on install?

eyberg:
yeh you're right it does - for some reason I didn't think it did

Issue #196, opened 2019-02-12
writes to /tmp/resolv and should build image directly


billhuey:
```billh@dungeon:~$ ops run hello-world
Downloading.. .staging/mkfs
 271.91 KiB / 271.91 KiB [==================================================================================================================] 100.00% 5.17 MiB/s 0s
Downloading.. .staging/boot.img
 22.50 KiB / 22.50 KiB [====================================================================================================================] 100.00% 9.33 MiB/s 0s
Downloading.. .staging/stage3.img
 1.40 MiB / 1.40 MiB [======================================================================================================================] 100.00% 8.84 MiB/s 0s
panic: open /tmp/resolv: permission denied

goroutine 1 [running]:
github.com/nanovms/ops/lepton.addDNSConfig(0xc00062e060, 0xc000135900)
        /home/billh/go/src/github.com/nanovms/ops/lepton/image.go:55 +0x27c
github.com/nanovms/ops/lepton.addFromConfig(0xc00062e060, 0xc000135900)
        /home/billh/go/src/github.com/nanovms/ops/lepton/image.go:119 +0x106
github.com/nanovms/ops/lepton.BuildManifest(0xc000135900, 0x5, 0x8cc033, 0x4)
        /home/billh/go/src/github.com/nanovms/ops/lepton/image.go:149 +0x1a9
github.com/nanovms/ops/lepton.BuildImage(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/billh/go/src/github.com/nanovms/ops/lepton/image.go:25 +0x2f
main.buildImages(0xc0001529a0)
        /home/billh/go/src/github.com/nanovms/ops/ops.go:143 +0xb8
main.runCommandHandler(0xc00011d180, 0xc0000db110, 0x1, 0x1)
        /home/billh/go/src/github.com/nanovms/ops/ops.go:113 +0x5db
github.com/spf13/cobra.(*Command).execute(0xc00011d180, 0xc0000db0e0, 0x1, 0x1, 0xc00011d180, 0xc0000db0e0)
        /home/billh/go/src/github.com/spf13/cobra/command.go:766 +0x2cc
github.com/spf13/cobra.(*Command).ExecuteC(0xc00018cc80, 0x8, 0x0, 0xc00018c780)
        /home/billh/go/src/github.com/spf13/cobra/command.go:852 +0x2fd
[...]

Issue #147, opened 2019-01-16
rm option for ops run


cbarraford:
Similar to how docker has "docker run --rm", we should have a similar option for ops to cleanup and remove any elements it may have created during a run. This may include the image file as well as the `.staging` directory

eyberg:
this behavior has changed and only one image is kept in a single repo

PR #208, opened 2019-02-14
Start work on removing .staging directory

anikitenko:
This PR created to change the way ".staging" directory was used.
In this PR:
* temp directory is determined based on OS you are running
* directory in "temp" one is named according to cons `OpsDir`
* file that contains information about contents in that directory is const `TMPContentsFile`
* another temp directories are created using Go `ioutil.TempDir` with prefix const `TMPStagingDirectoryPrefix`
* Uniqueness is done using the following methods:
  * If user uses `run` command then full path to it's binary is used (ex. `/Users/alexvwan/dev/src/test/server`) with it's association to new unique temp directory
  * If user uses `load` command then full path to current directory + gathered from manifest `manifest.Program`


anikitenko:
So, in other words, instead of using ".staging" in each directory, now everything is located at: `<system temp dir>/<OpsDir>/*`. And information of existence of anything is located at `<system temp dir>/<OpsDir>/<TMPContentsFile>`

anikitenko:
BTW, such logic of defining uniqueness is only valid until user don't specify parameters related to location of boot images.. I thought that if user specify any than responsibility exactly on users's side

eyberg:
is there any reason besides debugging purposes that we even want to keep a .staging directory @tijoytom  && @mkhon ?? like - once the image is built imo we should just nuke it right? keep the resulting image but not the tmp stuff? and we could store the built image into something like ~/.ops/projects/my.img or something like that to help facilitate:

a) running it from anywhere (eg: outside the project root)

b) having a known good location to drop them in

also -- it doesn't need to be addressed in this pr but imo https://github.com/nanovms/ops/issues/196 prob. should go in the .staging (whatever becomes of this)

mkhon:
Can we keep mkfs, boot.img and stage3.img in a fixed location (e.g.
somewhere under ~/.ops)?

eyberg:
Id strongly prefer that as it can be reused

anikitenko:
@mkhon I have open PR to keep mkfs, boot.img and stage3.img in a fixed location #195 but work in progress there..

lepton/const.go
```diff
@@ -21,26 +21,34 @@ const manifest string = `(
...
+const TMPContentsFile string = "contents.json"
+const TMPStagingDirectoryPrefix string = "staging_"
+
 const ReleaseBaseUrl string = "https://storage.googleapis.com/nanos/release/%v"
 const DevBaseUrl string = "https://storage.googleapis.com/nanos/%v"
 
 const PackageBaseURL string = "https://storage.googleapis.com/packagehub/%v"
 const PackageManifestURL string = "https://storage.googleapis.com/packagehub/manifest.json"
 const PackageManifestFileName string = "manifest.json"
+const PackageRootPath string = "sysroot"
```
tijoytom: This should not be a global const if it's used only in one function.

lepton/image.go
```diff
@@ -74,23 +74,36 @@ func addHostName(m *Manifest, c *Config) {
 }
 
 func addFilesFromPackage(packagepath string, m *Manifest) {
```
tijoytom: I don't think these changes work, i might be mistaken. Also there is no point in traversing the directory twice.

lepton/config.go
```diff
@@ -19,6 +19,7 @@ type Config struct {
 	RunConfig    RunConfig
 	Force        bool
 	TargetRoot   string
+	LocalTMPDir  string
```
tijoytom: TMP  should not be a config.

lepton/image.go
```diff
@@ -204,17 +214,26 @@ func addMappedFiles(src string, dest string, m *Manifest) {
...
+
 	//  prepare manifest file
-	var elfmanifest string
+	elfmanifest := m.String()
 
-	elfmanifest = m.String()
 	// invoke mkfs to create the filesystem ie kernel + elf image
-	createFile(mergedImg)
-	args := []string{}
+	mergedImgPath := filepath.Join(c.LocalTMPDir, mergedImg)
```
tijoytom: Why is this necessary? creating and closing the temp file. Any way it's going to be created.

lepton/image.go
```diff
@@ -230,19 +249,35 @@ func buildImage(c *Config, m *Manifest) error {
 		return errors.Wrap(err, 1)
 	}
 
+	bootImageOpen, err := os.Open(c.Boot)
```
tijoytom: Again why do you want to open this file?

lepton/image.go
```diff
@@ -230,19 +249,35 @@ func buildImage(c *Config, m *Manifest) error {
...
+	if err != nil {
+		return errors.Wrap(err, 1)
+	}
+	defer bootImageOpen.Close()
+
+	mergedImgPathOpen, err := os.Open(mergedImgPath)
```
tijoytom: same as above

lepton/config.go
anikitenko: Okay, I'll remove this from config

lepton/const.go
anikitenko: Hm.. I thought it's better to move "sysroot" to const instead of hardcoding. If you think it's not needed, I'll remove it

lepton/image.go
anikitenko: I tested these changes and they are working. Also I'm not traversing directory twice: first `filepath.Walk` traversing "sysroot", second one is it's parent directory ignoring "sysroot".

lepton/image.go
anikitenko: I thought that there is no need to keep it open...

lepton/image.go
anikitenko: `c.Boot` and `mergedImgPath` are need to be open to perform `io.Copy` to final image. I thought it's better to rewrite code to use `io.Copy` instead of running shell command `cat`

lepton/image.go
anikitenko: ah, got what you mean.. I'll remove `mergedImg.Close()`

lepton/image.go
tijoytom: is there a problem with cat? if not  i would keep it as such, using io.Copy buys you nothing, it introduce more code while solving a non-existing problem.

lepton/const.go
tijoytom: I am fine if you want to keep it a const but then don't export it.(ie uppercase P)

lepton/image.go
anikitenko: Actually there is no problem with `cat` but `io.Copy` is more a "Go way" of working with binary data..

lepton/const.go
anikitenko: ah, okay, I'll make it not exported

lepton/image.go
mkhon: I would prefer io.Copy as dependencies on external programs are bad

eyberg: yeh - it might be a bit slower but agreed - we should have as little exec.command as possible

anikitenko:
I made changes based on discussions. I noticed that "image" is now placed in "$HOME/.ops/images" but it's name is hardcoded to "image". But what will happen if I run 2 images? I thought it's better to place it in temp directory determined by `GetTMPPathByProgramPath` func.. but for now I'm not exactly sure on how to do this.. :(

anikitenko:
Just now an idea came to my mind.. when user has "shared" account on some server, some administrators can restrict access to `/tmp` directory.. What do you think if I rewrite `GetTMPPathByProgramPath` func to set path to staging like the following: `filepath.Join(GetOpsHome(), "staging")`? And the other functionality of func will remain? I think that this could resolve problem #196 and will implement in #210

eyberg:
the tmp. image we don't want to keep around for anything other than debug purposes but for built images we need to store somewhere so it can be re-used again - in a known place - the images do need to be named in order for this to work but that wasn't clearly communicated - i suppose for now we can allow the old behavior by naming the image by default whatever directory it is in like go does but there should be a name argument && || spot in the config.json that will name it properly

sorry for the confusion

anikitenko:
Ah, I see... work on that could be next step from moving from ".staging" directory.

anikitenko:
Updated pull request to include fix of #217

anikitenko:
Also in these changes I changed `os.TempDir` to `GetOpsHome()` in order to make sure that errors like #196 won't happen

And also in these changes I added funcs `RemoveTMPPathByProgramPath` and `RemoveImagePathByProgramPath` so they can be used to implement #147

lepton/image.go
```diff
@@ -225,19 +241,43 @@ func buildImage(c *Config, m *Manifest) error {
...
 		return errors.Wrap(err, 1)
 	}
-	catcmd.Wait()
+
+	if err := mergedImg.Close(); err != nil {
+		return errors.Wrap(err, 1)
+	}
+
+	if len(c.Debugflags) == 0 {
+		if err := RemoveTMPPathByProgramPath(programAbsPath); err != nil {
```
tijoytom: Not sure i understand this, how is removing temp related to debug flags?

anikitenko: Ah, also forgot to mention that I added the following:
* If we have debug option specified than tempimage will remain
* If nothing debug specified that tempimage is removed along with it's temp directory at `$HOME/.ops/<temp_path>`

anikitenko: As @eyberg said that `tempimage` is more for debug purposes...

tijoytom: Sorry for the confusion, tempimage is an artifact of how we build the images. First we merge kernel an user executable to a file (which is the tempimage), then we merge tempimage and the boot loader to final image. So you can remove the temp-image right after the final merge.

anikitenko: I see.. but again, for debugging purposes maybe we can leave it if user specifies debug option?

eyberg: i don't see a need for tempimage or any of the other artifacts to be present -- '-d' works fine w/the resulting built image - for instance in the old model -- 'image' we want to keep but anything in ~/.staging we don't want to keep

anikitenko: Actually tempimage is not stored in .staging, everything is under ~/.ops, but Only in case someone specified -d option, tempimage is saved, in other cases it’s removed

eyberg: tempimage does not need to be saved at all -- '-d' only uses the built image - we can completely remove tempimage

anikitenko: Ah, okay, I’ll update PR to always remove it

anikitenko: Updated PR to always remove tempimage

lepton/images.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil
+	}
+
+	// Check if file which describes contents of images data exists
+	if _, err := os.Stat(imagesOpsFilePath); os.IsNotExist(err) {
+		return nil
+	}
+
+	// Read contents of file which describes contents of images data
+	contentJSON, err := ioutil.ReadFile(imagesOpsFilePath)
```
tijoytom: Not sure i understand why all this is necessary, are you storing the list of temp files in a json and using it to delete them later? Why can't you just delete the temp folder ie `os.RemoveAll(temp)` would delete everything.

anikitenko:
Closing because PR is outdated

PR #215, opened 2019-02-17
option to skipbuild and only run

tijoytom:


tijoytom:
If you don't want to build again, but just run an existing image.

Issue #141, opened 2019-01-15
intel hax should be installed by default for mac users


eyberg:
there are pre-made verified builds by intel

one note - it looks like only osx 10.10 and above is supported so there can be a quick version check done to see if we can even install or not but if we can we should

ExternalReality:
@eyberg can you point me at the pre-made builds. For some reason I am having trouble finding them.

eyberg:
just copying some old notes -->

https://github.com/intel/haxm
https://software.intel.com/en-us/blogs/2014/10/20/how-to-ensure-you-are-using-intel-haxm-for-android-emulator


```
If you want to enable HAX (intel hardware acceleration) on Mac you have to options:

You can either roll your own kernel extension or you can download a dmg:

Roll Your Own Extension
boot into recovery mode (reboot && cmd-shift-R) and disable system integrity either of the two following methods:
csrutil enable --without kext
csrutil disable
git clone https://github.com/intel/haxm

build the kernel extension: (note: substitute the 10.13 for whatever is in sw_vers -productVersion)

xcodebuild -config Release -sdk macosx10.13

sudo chown -R root:wheel intelhaxm.kext

sudo kextutil intelhaxm.kext

Download a signed dmg
https://software.intel.com/en-us/android/articles/installation-instructions-for-intel-hardware-accelerated-execution-manager-mac-os-x

You should now be able to see that hax is working: qemu-system-x86_64 -accel hax
```

this task is actually composed of a few sub-tasks imo:

- [ ] enumerate osx version to determine what .dmg should be loaded 
- [ ] prompt but don't demand the user install hax
- [ ] need perms request in installer
- [ ] possibly create new signed dmgs 
- [ ] implement kexst check for osx (to see if there is hax enabled)
- [ ] re-eval perms issues to make this easy for end-users - i'm kinda thinking that if the installer script can be smart enough to detect vt-x and friends it should just go ahead and setup necessary permissions so that the user can run this stuff w/out sudo everytime - there's an implication here that most real deployments are going to be pushed through public clouds regardless so it's a bit of a mute issue for local

eyberg:
just as a fyi - we should be disabling sip for end-users - that's just a debugging mechanism

ExternalReality:
Some thoughs:

> enumerate osx version to determine what .dmg should be loaded

[...]

PR #221, opened 2019-02-19
Auto install haxm.

ExternalReality:
Things to do: https://github.com/nanovms/ops/issues/141

- [x] ~enumerate osx version to determine what .dmg should be loaded~
- [x] prompt but don't demand the user install hax
- [x] need perms request in installer
- [ ] possibly create new signed dmgs
- [x] implement kexst check for osx (to see if there is hax enabled)
- [ ] re-eval perms issues to make this easy for end-users - i'm kinda thinking that if the installer script can be smart enough to detect vt-x and friends it should just go ahead and setup necessary permissions so that the user can run this stuff w/out sudo everytime - there's an implication here that most real deployments are going to be pushed through public clouds regardless so it's a bit of a mute issue for local
- [x] use acceleration flag appropriate to host OS.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -208,10 +208,28 @@ ops_install_haxm() {
     # installed or if the haxm kext was otherwise loaded. 
     existing_install=`pkgutil --pkgs | grep haxm`
     loaded_module=`kextstat | grep intelhaxm`
-    if [ ! -z $existing_install ] || [ ! -z $loaded_module ]; then
+    if [ ! -z "$existing_install" ] || [ ! -z $loaded_module ]; then
         return
     fi
 
+    while true; do
+	prompt="Would you like to install Intel HAXM for hypervisor hardware acceleration support [yes/NO]? "
+	read -p "$prompt"
+	reply=`echo $REPLY | tr '[:upper:]' '[:lower:]'`
+
+	if [ -z "$reply" ]; then
+	    reply="no"
+	fi
+
+	case "$reply" in
+	    no) return
+		;;
+	    yes) break
+		;;
+	    *) echo "Valid responses are either \"yes\" or \"no\"".
+	esac
+     done
+    
     # get the binary package. No need to check the md5 since the we are going
     # for the included .dmg which is signed.
     curl -LJO https://github.com/intel/haxm/releases/download/v7.4.1/haxm-macosx_v7_4_1.zip
--- a/install.sh
+++ b/install.sh
@@ -206,14 +206,14 @@ ops_install_haxm() {
 
     # check to see if a version of the package was previously
     # installed or if the haxm kext was otherwise loaded. 
-    existing_install=`pkgutil --pkgs | grep haxm`
-    loaded_module=`kextstat | grep intelhaxm`
+    readonly existing_install=`pkgutil --pkgs | grep haxm`
+    readonly loaded_module=`kextstat | grep intelhaxm`
     if [ ! -z "$existing_install" ] || [ ! -z $loaded_module ]; then
         return
     fi
 
     while true; do
-	prompt="Would you like to install Intel HAXM for hypervisor hardware acceleration support [yes/NO]? "
+	readonly prompt="Would you like to install Intel HAXM for hypervisor hardware acceleration support [yes/NO]? "
 	read -p "$prompt"
 	reply=`echo $REPLY | tr '[:upper:]' '[:lower:]'`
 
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -199,6 +199,28 @@ ops_install_qemu() {
...
+ops_install_haxm() {
+    if [ "$OS" != "darwin" ]; then
+        return
+    fi
+
+    # check to see if a version of the package was previously
+    # installed or if the haxm kext was otherwise loaded. 
+    existing_install=`pkgutil --pkgs | grep haxm`
+    loaded_module=`kextstat | grep intelhaxm`
+    if [ ! -z $existing_install ] || [ ! -z $loaded_module ]; then
```
ExternalReality: The .mpkg in the .dmg does this check and also runs a script that checks to see if a binary is running (amongst other things). We do a quick check here to avoid having to go through the trouble of mounting the .dmg and such if we know the installer is just going to fail.

install.sh
```diff
@@ -199,6 +199,28 @@ ops_install_qemu() {
...
+        return
+    fi
+
+    # get the binary package. No need to check the md5 since the we are going
+    # for the included .dmg which is signed.
+    curl -LJO https://github.com/intel/haxm/releases/download/v7.4.1/haxm-macosx_v7_4_1.zip
+    tar -xvf haxm-macosx_v7_4_1.zip -C /tmp
+    hdiutil attach /tmp/IntelHAXM_7.4.1.dmg
+    installer -package /Volumes/IntelHAXM_7.4.1/IntelHAXM_7.4.1.mpkg -target /
```
ExternalReality: This requires permissions. It will fail as is.

ExternalReality:
https://github.com/nanovms/ops/issues/141

ExternalReality:
> implement kexst check for osx (to see if there is HAX enabled)

As of now we are doing this in the script, the installer also does this and a bit more - such as checking if the binary is running by mounting the .dmg, extracting a check script from the package, and running it. In short, the script will fail to install if the kext is loaded, the binary is running, or the package is installed.

ExternalReality:
If you want to uninstall the package (for example if you want to test the `install.sh` script again) it seems that you have to do two things:

1. Run the uninstaller found in the `.dmg` image.
2.  [Uninstall](https://wincent.com/wiki/Uninstalling_packages_(.pkg_files)_on_Mac_OS_X) the `.mpkg` found in the `.dmg` by hand.

If both of these tasks are not done, subsequent runs of the `ops` installer will skip trying to install `HAXM` (determining that `HAXM` is already installed). If the `ops` installed would allow the `HAXM` installer to run in the absence of performing those two steps, the `HAXM` installer itself would fail.

install.sh
eyberg: this is the part where a prompt to the user to indicate that they can get hardware acceleration if they sudo install this and letting the user gracefully opt-out if they so desire

eyberg:
yeh - it's not thought out much yet - but essentially we'd want to do that kexst check before doing any kvm operations as-is on osx - that or gracefully fail if kvm is enabled && it's not installed

ExternalReality:
@eyberg @tijoytom  - You needs [at least version 2.9.0 ](https://www.qemu.org/2017/11/22/haxm-usage-windows/) of qemu if you want to use HAXM acceleration. Should I do the version check or does ops support only qemu > 2.9.

tijoytom:
ops support qemu 2.5, do the version check and prompt the user to update qemu

eyberg:
for me I see -->

```
brew info qemu
qemu: stable 3.1.0 (bottled), HEAD
x86 and PowerPC Emulator
https://www.qemu.org/
/usr/local/Cellar/qemu/2.4.0.1 (123 files, 104.9MB)
  Poured from bottle on 2015-09-24 at 12:00:00
/usr/local/Cellar/qemu/2.10.0 (131 files, 156MB)
  Poured from bottle on 2017-10-02 at 21:41:05
/usr/local/Cellar/qemu/3.0.0 (137 files, 261.8MB) *
  Poured from bottle on 2018-09-21 at 15:08:33
```

I'd imagine most mac users are not used to using qemu and thus they should prob. have 3.x if they try through brew so -- if you want to put a warning that's fine but I think it'll be fine w/out

ExternalReality:
I am receiving some some issue when running the examples with `-accel hax` however the [macOS native hypervisor](https://developer.apple.com/documentation/hypervisor) accelerator looks to work fine out of the box using qemu options `-accel=hvf -cpu=host` with `ops`.

eyberg:
really? that'd be great if we can just use the native hypervisor instead

krytarowski:
HAXM can be also used on NetBSD (and Windows or Linux)..

eyberg:
we can have support for both - just an fyi - there's planned work to support many other hypervisors (and platforms)

ExternalReality:
Does the `Ops` installer still want to ask about `HAXM` on `Darwin`?

Maybe the installer could say something like:

`For hardware acceleration Ops uses hypervisor.framework by default; would you like to install Intel HAXM and use that instead?`.

Another option would be to add an option to the `ops run/load` Config and pass it down to `Qemu`. In this case the user is left to install `HAXM` herself.

eyberg:
i think options are nice but defaulting to hypervisor.framework is prob. a good default too

eyberg:
are you going to keep this one w/https://github.com/nanovms/ops/pull/230 or should we close it?

PR #222, opened 2019-02-19
add usr/lib to search path

tijoytom:


lepton/elf.go
```diff
@@ -86,7 +87,7 @@ func _getSharedLibs(targetRoot string, path string) ([]string, error) {
 			return nil, err
 		}
 	}
-	dt_runpath = append(dt_runpath, "/lib64", "/lib/x86_64-linux-gnu", "/usr/lib64", "/usr/lib/x86_64-linux-gnu")
+	dt_runpath = append(dt_runpath, "/lib64", "/lib/x86_64-linux-gnu", "/usr/lib", "/usr/lib64", "/usr/lib/x86_64-linux-gnu")
```
mkhon: @tijoytom out of curiosity - which Linux distro uses /usr/lib and do you think /lib sbould also be added.

tijoytom: It's Ubuntu 16.4, I think it's not about the distribution, but many libs use symbolic links in /use/lib pointing to correct version

PR #224, opened 2019-02-20
generate image names

tijoytom:


tijoytom:
@eyberg 
if user supply an imagename "myimage", the final image will be $HOME/.ops/images/myimage
if no imagename is supplied, final image will be $HOME/.ops/<program_mmddhhmmss.img>  where mmddhhmmss is month-day-hour-minute-second

eyberg:
the only feedback I'd say is rm the timestamp for this

PR #226, opened 2019-02-20
copy all userconfig to package config

tijoytom:


tijoytom:
@wjhun this's why your local config it not work for package load.

PR #230, opened 2019-02-23
Add accel hvf flag

ExternalReality:
Here we implement a virtualization hardware support check by making the system call on `Darwin` and by reading `/proc/cpuinfo` on `Linux`. We use this to know if we should add an `-accel` option to qemu. In neither case do we shell. 

However with this I have noticed that, while ` qemu  -accel-hvf ` works fine on `macos`; when using `-accel kvm` on `linux` qemu hangs when running an ops image. I've tested this with a `helloworld.go` image on my `AMD Ryzen` host. I'll certainly test  some more on other machine archs and distributions. 

TODO:

- [x] Have installer prompt to add user to `kvm` group on linux.
- [x] Do a `qemu` version check to know what flags/options to add.
- [x] Do User group check before adding flag
- [x] Find out why adding `-accel kvm/-enable kvm` causes the ops generated `qemu` command to hang. ~(I've tried this command with a minix3 image and it also hangs so its probably qemu and not the ops image - I'm digging through bug reports and docs - some old bug reports say that the command hangs when adding -enable-kvm but also report that the problem has been fixed)~

### DIFF
--- a/lepton/hv_support_linux.go
+++ b/lepton/hv_support_linux.go
@@ -1,0 +1,37 @@
+package lepton
+
+import (
+	"io/ioutil"
+	"strings"
+)
+
+func hv_support() (bool, error) {
+	const intel string = "vmx"
+	const amd string = "svm"
+
+	b, err := ioutil.ReadFile("/proc/cpuinfo")
+	if err != nil {
+		return false, err
+	}
+
+	content := string(b)
+	lines := strings.Split(content, "\n")
+
+	for _, line := range lines {
+		kvp := strings.Split(line, ":")
+		if len(kvp) == 0 {
+			continue
+		}
+		if strings.TrimSpace(kvp[0]) != "flags" {
+			continue
+		}
+		for _, flag := range strings.Split(kvp[1], " ") {
+			if flag == intel || flag == amd {
+				return true, nil
+			}
+			continue
+		}
+	}
+
+	return false, nil
+}
--- a/lepton/qemu.go
+++ b/lepton/qemu.go
@@ -6,6 +6,10 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"os/user"
+	"regexp"
+	"runtime"
+	"strconv"
 	"strings"
 	"syscall"
 )
@@ -219,6 +223,84 @@ func generateMac() string {
 		octets[0], octets[1], octets[2], octets[3], octets[4], octets[5])
 }
 
+func qemuVersion() (string, error) {
+	versionData, err := exec.Command(qemuBaseCommand, "--version").Output()
+	if err != nil {
+		return "", err
+	}
+	return parseQemuVersion(versionData), nil
+}
+
+func parseQemuVersion(data []byte) string {
+	rgx := regexp.MustCompile("\\b[0-9]\\.[0-9]\\.[0-9]\\b")
+	return string(rgx.Find(data)[:3])
+}
+
+// kvmGroupPermission returns true if the current user is part of the kvm qroup,
+// false if she is not, or an error. or an error.
+func kvmGroupPermissions() (bool, error) {
+	return groupPermissions("kvm")
+}
+
+// groupPermissions returns true if the user is in the group specified by the
+// argument string, false if she is not, or an error.
+func groupPermissions(groupName string) (bool, error) {
+	currentUser, err := user.Current()
+	if err != nil {
+		return false, err
+	}
+	gids, err := currentUser.GroupIds()
+	if err != nil {
+		return false, err
+	}
+	g, err := user.LookupGroup(groupName)
+	if err != nil {
+		return false, err
+	}
+	for _, gid := range gids {
+		if g.Gid == gid {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// versionCompare compares Qemu version numbers. If the the first argument is
+// greater then true is returned, if the second argument is greater
+// then versionCompare returns false, otherwise it returns true.
+func (q *qemu) versionCompare(v1, v2 string) (bool, error) {
+	ver1 := strings.Split(v1, ".")
+	ver2 := strings.Split(v2, ".")
+	const formatError = "improperly formated qemu version %q"
+	errV1 := fmt.Errorf(formatError, v1)
+	errV2 := fmt.Errorf(formatError, v2)
+
+	if len(ver1) < 2 {
+		return false, errV1
+	}
+	if len(ver2) < 2 {
+		return false, errV2
+	}
+	majorV1, err := strconv.Atoi(ver1[0])
+	minorV1, err := strconv.Atoi(ver1[1])
+	if err != nil {
+		return false, errV1
+	}
+	majorV2, err := strconv.Atoi(ver2[0])
+	minorV2, err := strconv.Atoi(ver2[1])
+	if err != nil {
+		return false, errV2
+	}
+	if majorV1 < majorV2 {
+		return false, nil
+	}
+	if majorV1 == majorV2 && minorV1 < minorV2 {
+		return false, nil
+	}
+	return true, nil
+}
+
 func (q *qemu) addFlag(flag string) {
 	q.flags = append(q.flags, flag)
 }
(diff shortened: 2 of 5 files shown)


lepton/hv_support_linux.go
```diff
@@ -0,0 +1,37 @@
+package lepton
+
+import (
+	"io/ioutil"
+	"strings"
+)
+
+func hv_support() (bool, error) {
+	const intel string = "vmv"
```
eyberg: vmx

lepton/qemu.go
```diff
@@ -228,6 +229,22 @@ func (q *qemu) addOption(flag, value string) {
...
+	// implemented, then errors in detecting hardware acceleration support
+	// should be reported. Here we treat any error in detecting hardware
+	// acceleration the same as detecting no hardware acceleration - we don't
+	// add the flag.
+	if ok, err := hv_support(); ok && err == nil {
+		if runtime.GOOS == "darwin" {
+			q.addOption("-accel", "hvf")
+			q.addOption("-cpu", "host")
+		} else {
+			q.addOption("-accel", "kvm")
```
eyberg: which qemu is this? on both qemu 3.x on mac and 2.5 on ubuntu I don't see accel flags

lepton/qemu.go
```diff
@@ -236,8 +253,8 @@ func (q *qemu) setConfig(rconfig *RunConfig) {
 	if rconfig.Bridged {
 		devType = "tap"
 		ifaceName = rconfig.TapName
-		q.addFlag("-enable-kvm")
 	}
+	q.addAccel()
```
eyberg: as much as I'd like to this prob. shouldn't be default w/out having a better install exp to setup permissions correctly - your average user won't know that this takes permissions to run and the average box is not setup to allow access to kvm as non-root

eyberg: idk if macos cares about that or not w/their native hypervisor framework - if not then totally allow it for default on mac

lepton/hv_support_linux.go
ExternalReality: :-(

lepton/qemu.go
ExternalReality: `macos` doesn't seem to need additional added user permissions; AFACT  `qemu` from `brew` is not sandboxed and thus doesn't need the added entitlement. Therefore nothing extra is required from the user for the `-accel hvf -cpu host` flags to be added. 

For the other supported systems the install can add the user to the `kvm`. Ops can also skip adding the `-accel kvm` flag if it finds it was run by a user that is not in the `kvm` group and perhaps issue a message notifying the user to add herself to the group.

lepton/qemu.go
ExternalReality: As far as I can tell from the change logs the [-accel flag is 3.0+](https://wiki.qemu.org/ChangeLog/3.0). On qemu 3.1.0 on `macos` I see the flags listed on the man page. 

There can be a version check to choose the correct option/flag to avoid any deprecation warnings.

ExternalReality: Is should be noted that the default `qemu` from apt on `debian 9 stretch` is 2.8.

ExternalReality: `-accel` seems to be 2.12+ not 3.0+, earlier change logs do not mention.

lepton/qemu.go
eyberg: we'll need to add the group functionality to the install script first before putting this in

lepton/qemu.go
```diff
@@ -228,6 +310,44 @@ func (q *qemu) addOption(flag, value string) {
...
+	}
+	if runtime.GOOS == "linux" {
+		ok, err := kvmGroupPermissions()
+		if !(ok && err == nil) {
+			return
+		}
+		if ok, _ := q.versionCompare(qemuVersion, accelFlagVersion); ok {
+			q.addOption("-accel", "kvm")
+		} else {
+			q.addFlag("-enable-kvm")
```
tijoytom: The argument accel=kvm of the -machine option is equivalent to the -enable-kvm option. Just use `enable-kvm` and you can get rid of the version check for linux too.
https://wiki.archlinux.org/index.php/QEMU

lepton/qemu.go
tijoytom: accel=kvm and -enable-kvm are equivalent,  just add `-enable-kvm` and remove the linux version check since `-enable-kvm` works across all versions.

ExternalReality: @tijoytom Will do, I have been trying to run things with `kvm` enabled (by any option) with no luck. The command hangs. Digging through mailing lists and forums reveals some bugs that inidicate that the "-accel kvm/-enable-kvm" options can hang and yet none of the proposed solutions worked for me. I've tried different kernels and different distros and still no luck. 

Would be you able to confirm this by `ops run -v <binary>` and they by sticking an `-enable-kvm` at the end of the command?

eyberg: i was going to respond to this - i'm pretty sure you have to set the acclerator used on osx ->

```
-accel hax
HAX is working and emulator runs in fast virt mode.
qemu-system-x86_64: terminating on signal 2 from pid 24570 (<unknown process>)
```

since kvm is linux specific it makes sense you have to set the appropriate flag here

lepton/qemu.go
ExternalReality: Yes, the group functionality was added. If the installer detects that the current user is not in the `kvm` group it asks to add the user. If "yes" is chosed, it will notify the user that it is using `sudo` to add him to the `kvm` group. If successful in adding the user the installer notifies the user that the changes will take affect after a subsequent login. This is implemented by the `sh` function [op_user_permissions](https://github.com/nanovms/ops/blob/d7fba40ffc1e600f16ab8569d2aa94dfd2aca318/install.sh#L204-L251) in `install.sh`.

lepton/qemu.go
tijoytom: what's  your qemu version?

tijoytom: I see the same problem, i think it might have to do something with qemu version, i know for sure 2.5 worked. I dont have qemu2.5 now, if you have it ready can you please try it.

ExternalReality: I've tried on `2.8.1` which is the default on `Debian 9 stretch` and I have tried on `3.1.0` which is the default on `Manjaro` (I know manjaro is not supported I just tried it anyway since I have it available). I have also tried it on `3.1.0` on `Debian 9` and `3.1.0` on `Fedora29`.

ExternalReality: Actually that is `3.0.0` on `fedora29`.

ExternalReality: @tijoytom You were correct. `-enable-kvm`  works fine on `2.5.0`. In fact it works fine on `2.5.0`, `2.6.0`... all the way to `2.8.0` where the ops generated command stops working.

For whatever reason, after qemu-version `2.7.1`, you must disable hardware `irq` handling and allow qemu to continue to fully emulate `irq` when using `-enable-kvm`.

So AFACT:

**When using the flag `-enable-kvm` we must also use `-machine kernel-irqchip=off`**

Unless there is some other configuration which allows kvm with kernel-irqchip that I am not seeing. The qemu docs aren't the greatest in this area.

ExternalReality:
Alright this should be OK now:

1) Prompts the user to be added to `kvm` group (if not in group) on install and notifies user they need to re-log.
2) Performs qemu version check for mac (hvf) support 
3) Avoids hanging `-enable-kvm` for qemu version > 2.7.1 by adding turning off hardware irq
4) Checks for hardware (VT-x/AMD-V) support before adding any acceleration flag

lepton/qemu.go
eyberg: need to let @wjhun && @mkhon chime in here - I think this is the msi vs. msi-x problem you're hitting https://github.com/nanovms/nanos/pull/172 && https://github.com/nanovms/nanos/issues/27 - kernel-irqchip=off prob. is not the fix we want to do here

ExternalReality: I see.

ExternalReality: I wonder what the status of this is because with most of the default supported versions of `qemu` (those that are installed by default on supported distros) if a user chooses the `--bridged` option it enables `kvm` and qemu will hang. The `-kernel-irqchip=off` option may not be the optimal solution, but it does at least work and basic `time` benchmarks show that the image is executing significantly faster than without `--enable-kvm`.

eyberg: another option to look at (cause we've already reported this problem upstream) is this ->

I haven't personally looked at the results between your solution and this but it'd be interesting to get some stats on it

```
-global driver=virtio-blk-pci,property=ioeventfd,value=off
```

the other thing is that I'm not opposed to turning this on selectively right now for the affected qemu versions as long as we let the others remain w/out it

lastly - we are in all likelihood converting to msi-x in the very near future which kinda obviates this entire issue - so this would be more of a thing if you want to spend time on it now or just wait

eyberg:
@ExternalReality sorry for the lateness on this - feel free to fix conflicts and merge - your option is fine for now

lepton/qemu.go
```diff
@@ -228,18 +310,52 @@ func (q *qemu) addOption(flag, value string) {
...
+			return
+		}
+	}
+	if runtime.GOOS == "linux" {
+		ok, err := kvmGroupPermissions()
+		if !(ok && err == nil) {
+			return
+		}
+		q.addFlag("-enable-kvm")
+		q.addOption("-machine", "kernel-irqchip=off")
```
mkhon: kernel-irqchip=off will no longer be required - fixed in nanovms/nanos#587

ExternalReality:
I'd like to test this more against the default versions of `qemu` on supported distributions. While I confident that the code does as intended, I simply want to be sure that the `--enable-kvm` behaves with ops images on the versions of qemu installed by default supported distributions (including perhaps a couple of releases back of those distributions or LTSs) .

ExternalReality:
I've tested with the below `qemu` versions with `nanos` nightly image on `Debian 9` with intel vmx. Like @mkhon mentioned the `kernel-irqchip=off` option is no longer needed with the `enable-kvm` flag - the image does not hang. Thank you. 

```
qemu-2.6.0/x86_64-softmmu/qemu-system-x86_64
qemu-2.7.0/x86_64-softmmu/qemu-system-x86_64
qemu-2.7.1/x86_64-softmmu/qemu-system-x86_64
qemu-2.8.0/x86_64-softmmu/qemu-system-x86_64
qemu-3.1.0/x86_64-softmmu/qemu-system-x86_64
```

ExternalReality:
@mkhon, removed the flag. Thanks for the fix. If you agree with that last commit (which removes the flag). I will squash and ask for a merge.

mkhon:
> @mkhon, removed the flag. Thanks for the fix. If you agree with that last commit (which removes the flag). I will squash and ask for a merge.

@ExternalReality yes, looks good to me

ExternalReality:
@tijoytom this one may be ready.

PR #237, opened 2019-02-28
move temp files to ~/.ops/common and also add passwd

tijoytom:


lepton/image.go
```diff
@@ -46,30 +46,47 @@ func createFile(filepath string) (*os.File, error) {
...
 func addDNSConfig(m *Manifest, c *Config) {
-	data := []byte("nameserver ")
-	data = append(data, []byte(c.NameServer)...)
-	temp := path.Join(os.TempDir(), "resolv")
-	err := ioutil.WriteFile(temp, data, 0666)
-	if err != nil {
-		panic(err)
+	common := GetOpsCommon()
+	resolv := path.Join(common, "resolv.conf")
+	if _, err := os.Stat(resolv); os.IsNotExist(err) {
```
eyberg: for these - i'd just make these tmp files - I don't see a reason to keep them around at all - also - they need to be image specific - not just for multi-tenants - the ~/.ops takes care of that but if you have multiple ones on the same system - eg: hostname for db would be diff. for hostname for the webapp

eyberg:
https://github.com/nanovms/nanos/issues/351

Issue #233, opened 2019-02-26
error gracefully on missing binary


eyberg:
I stupidly pasted this into the shell and of course it blew up cause there isn't a binary named server here - instead of throwing a panic/calltrace we should just gracefully error out and tell the user what the problem is - there's actually quite a few of these in the codebase that could use a human readable error response

```
➜  ~   NANOS_UPDATE=1 ops run server

Downloading.. /Users/eyberg/.ops/1.0/nanos-release-darwin-1.0.tar.gz
 1.05 MiB / 1.05 MiB [===============================================================================================================================================================] 100.00% 3.48 MiB/s 0s
Finding dependent shared libs
*os.PathError server: lstat server: no such file or directory
/Users/eyberg/go/src/github.com/nanovms/ops/lepton/elf.go:72 (0x12e8ac9)
        _getSharedLibs: return nil, errors.WrapPrefix(err, path, 0)
/Users/eyberg/go/src/github.com/nanovms/ops/lepton/elf.go:134 (0x12e961d)
        getSharedLibs: libs, err := _getSharedLibs(targetRoot, path)
/Users/eyberg/go/src/github.com/nanovms/ops/lepton/image.go:168 (0x12eae6b)
        BuildManifest: deps, err := getSharedLibs(c.TargetRoot, c.Program)
/Users/eyberg/go/src/github.com/nanovms/ops/lepton/image.go:24 (0x12e9d4f)
        BuildImage: m, err := BuildManifest(&c)
/Users/eyberg/go/src/github.com/nanovms/ops/ops.go:220 (0x1354922)
        buildImages: err = api.BuildImage(*c)
/Users/eyberg/go/src/github.com/nanovms/ops/ops.go:127 (0x1352cbb)
        runCommandHandler: buildImages(c)
/Users/eyberg/go/src/github.com/spf13/cobra/command.go:766 (0x1349d4c)
        (*Command).execute: c.Run(c, argWoFlags)
/Users/eyberg/go/src/github.com/spf13/cobra/command.go:852 (0x134a8ed)
        (*Command).ExecuteC: err = cmd.execute(flags)
/Users/eyberg/go/src/github.com/spf13/cobra/command.go:800 (0x134a5cb)
        (*Command).Execute: _, err := c.ExecuteC()
/Users/eyberg/go/src/github.com/nanovms/ops/ops.go:667 (0x1359b5c)
        main: rootCmd.Execute()
[...]

PR #239, opened 2019-02-28
Handle missing file error for run command.

ExternalReality:
https://github.com/nanovms/ops/issues/233

### DIFF
--- a/ops.go
+++ b/ops.go
@@ -566,12 +566,27 @@ func cmdListPackages(cmd *cobra.Command, args []string) {
 	table.Render()
 }
 
+func validateRunCmd(cmd *cobra.Command, args []string) error {
+	const MIN_ARGS = 1
+	if len(args) < MIN_ARGS {
+		return cobra.MinimumNArgs(MIN_ARGS)(cmd, args)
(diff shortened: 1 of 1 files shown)


ops.go
```diff
@@ -566,12 +566,27 @@ func cmdListPackages(cmd *cobra.Command, args []string) {
 	table.Render()
 }
 
+func validateRunCmd(cmd *cobra.Command, args []string) error {
```
ExternalReality: Only, I would recommend parsing the argument array into a structure before validation since matching the raw argument positions is error prone and hard to test. Structured input validation, IMHO, would be the first step in eliminating the panics. Second would be handling those errors that bubble up to the top with a logging package this ways errors can be easily formatted for stdout and written to file.

tijoytom:
This is already fixed, I forgot to close the bug. Please don't merge this one

tijoytom:
#235

tijoytom:
@ExternalReality I am going to close this one since it's already fixed.

ExternalReality:
@tijoytom No issue. thx.

PR #243, opened 2019-03-13
use ldd for linux

tijoytom:
Bring back ldd for linux, darwin still use manual method. There are couple of issues with current manual method. 
1. We don't use ld.config.so and on some linux distribution shared libs can be in  weird path. ldd seems to handle them.
2. There can be  more than one match, for example a program could be linked against 32 bit version of the shared lib, but we will end up picking the 64 bit version. 
There might be other issues too, these can be fixed , for example use ld.config.so, check the library if it's 32 or 64 bit etc..But still we might end up with difference with what ldd is doing. I would rather have, if ldd finds it we find it and just re-use ldd.


tijoytom:
@eyberg i am bringing back ldd for linux.

PR #244, opened 2019-03-14
change version check to allow downgrading

tijoytom:


tijoytom:
Change the version check to allow downgrade, now we can point latest to any version we want.

PR #246, opened 2019-03-15
make ops works even if we cant lookup the latest version

tijoytom:
Right now we seg fault, if there is a local build use it.

### DIFF
--- a/lepton/const.go
+++ b/lepton/const.go
@@ -144,10 +144,23 @@ func updateLocalRelease(version string) error {
 
 var LatestReleaseVersion string = getLatestRelVersion()
 
+const (
+	WarningColor = "\033[1;33m%s\033[0m"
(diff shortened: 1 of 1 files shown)


lepton/const.go
```diff
@@ -144,10 +144,23 @@ func updateLocalRelease(version string) error {
...
+	resp, err := http.Get(ReleaseBaseUrl + "latest.txt")
+	if err != nil {
+		fmt.Printf(WarningColor, "version lookup failed, using local.\n")
+		if LocalReleaseVersion == "0.0" {
+			fmt.Printf(ErrorColor, "No local build found.")
+			os.Exit(1)
+		}
+		return LocalReleaseVersion
+	}
 	data, _ := ioutil.ReadAll(resp.Body)
```
eyberg: i'd prefer to deal w/these but the rest is fine

PR #249, opened 2019-03-17
use virtio-blk as a drive type;any way this was the default before

tijoytom:


tijoytom:
This is supposed to fix https://github.com/nanovms/nanos/issues/594, but i still repro GPF with this change. I think the difference is i am using user mode networking with kvm-enabled while the `make run` in nanos is using tap. I need to look in to more details.

Issue #254, opened 2019-03-20
loading a missing pkg should tell the user no package exists rather than blowing up


eyberg:
this is another one of those dev-ux friendly type of things - if it's not in the manifest just let them know - maybe even ask the user if it should be?

```
➜  ~  ops load ruby_2.3.14234
Downloading.. /Users/eyberg/.ops/packages/ruby_2.3.14234.tar.gz
 224 B / 224 B [====================================================================================================================================================================] 100.00% 28.37 KiB/s 0s
Extracting /Users/eyberg/.ops/packages/ruby_2.3.14234.tar.gz to /Users/eyberg/.ops/.staging/ruby_2.3.14234
panic: gzip: invalid header

goroutine 1 [running]:
github.com/nanovms/ops/lepton.ExtractPackage(0xc000169c00, 0x31, 0xc000187980, 0x1b)
        /home/tijoytom/go/src/github.com/nanovms/ops/lepton/package.go:35 +0x472
main.loadCommandHandler(0xc000568a00, 0xc000286e90, 0x1, 0x1)
        /home/tijoytom/go/src/github.com/nanovms/ops/ops.go:393 +0x2d8
github.com/spf13/cobra.(*Command).execute(0xc000568a00, 0xc000286e60, 0x1, 0x1, 0xc000568a00, 0xc000286e60)
        /home/tijoytom/go/src/github.com/spf13/cobra/command.go:766 +0x2cc
github.com/spf13/cobra.(*Command).ExecuteC(0xc000568f00, 0x8, 0x0, 0xc000568a00)
        /home/tijoytom/go/src/github.com/spf13/cobra/command.go:852 +0x2fd
github.com/spf13/cobra.(*Command).Execute(0xc000568f00, 0xc0000d7ec0, 0x1)
        /home/tijoytom/go/src/github.com/spf13/cobra/command.go:800 +0x2b
main.main()
        /home/tijoytom/go/src/github.com/nanovms/ops/ops.go:689 +0x15bc
```

ExternalReality:
https://github.com/nanovms/ops/pull/270

eyberg:
thanks!

PR #270, opened 2019-03-26
Add ops get which simply downloads a package.

ExternalReality:
Command `ops get` to download a package from `ops list` and nothing more.


lepton/image.go
```diff
@@ -359,8 +359,7 @@ func DownloadReleaseImages(version string) error {
 }
 
 func DownloadFile(filepath string, url string, timeout int) error {
-
-	fmt.Println("Downloading..", filepath)
+	fmt.Println("Downloading..", url)
```
ExternalReality: Quick fix. It now says "Downloading.. <url>" instead of "Downloading.. <$OPS_HOME>"

ops.go
```diff
@@ -561,6 +561,10 @@ func loadCommandHandler(cmd *cobra.Command, args []string) {
 	hypervisor.Start(&c.RunConfig)
 }
 
+func getCommandHandler(cmd *cobra.Command, args []string) {
+	api.DownloadPackage(args[0])
```
ExternalReality: Error handling can be improved. The `DownloadPackage` command, if the argument package does not exists, will download an `xml` response from google storage whose contents claim `access denied` the command does not detect that is hasn't downloaded a package (but a xml file) and succeeds.

Not sure what the plans are for storing packages so I won't attempt to fix this without guidance.

eyberg: https://github.com/nanovms/ops/issues/254 - although i wouldn't spend too much time on it - we'll prob. be moving to a minio at some point in the future

ExternalReality: Cool. Checking the specified package against the package list would be a quickly implementable check and similar to how apt etc seem to work.

PR #269, opened 2019-03-25
Change config field name from UseKvm to Accel.

ExternalReality:

PR #272, opened 2019-03-26
Add accel(x) flag for cpu virt ext. support.

ExternalReality:
`Ops run -x <elf>`
`Ops load --accel <package> ...`

There is a little bit of an opportunity to refactor the `load` and `run` flags -> config stuff of course.


ExternalReality:
https://github.com/nanovms/ops/pull/269 will require change if this PR is merged first.

PR #279, opened 2019-03-30
"all" now depends on "deps" so they are automatically fetched

mkhon:


ExternalReality:
[Go modules](https://blog.golang.org/using-go-modules) simplify dependency management a great deal and are built into the Go platform. Auto scanning imports and the like on relevant commands. The feature is relatively new.

PR #280, opened 2019-03-30
Do not run TestGroupPermission on macOS

mkhon:
Not sure why this test is required. May be it should be removed altogether.

### DIFF
--- a/lepton/qemu_test.go
+++ b/lepton/qemu_test.go
@@ -2,7 +2,6 @@ package lepton
 
 import (
 	. "fmt"
-	"os/user"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


ExternalReality:
Lepton code tests for `kvm` group permissions before trying to add the `enable-kvm` flag to qemu. I added the test to ensure that the code was behaving properly. The code wasn't mean to run on MacOS where group permissions are not needed to add the relevant hypervisor flag (hvf). @mkhon like you mention the test could probably be removed, the test was a transient thing to see if all the checks ops does before enabling hypervisor support were working: `cpu support`, `kvm module installed`, `user member of kvm group`.

eyberg:
yeh - after looking at this this doesn't seem like something we should have as an automated test - i'd vote just to rm

PR #312, opened 2019-04-12
nanospkg command for easy building ops packages

skupr:
It's not a ready PR yet, it's more like a topic to dicuss

When creating packages for ops, I've noted that it's very mechanical process, and under the hood it's very similar to ops run command.
So I've created separate utility that does most common tasks to create package.

Build it by
`$ go build buildcmd/nanospkg`

Run example:
`$ nanospkg build -c config.json /usr/sbin/haproxy -v 1.8.8`
So it's pretty similar to 'ops run', the only addition is mandatory 'version' field.

Not sure if this should be a separate util or include it to ops commands, like 'ops pkg build', probably

Also, I've done a couple of hacks to ops code, to be able to run it from another package, not sure yet how to do better.

@eyberg What do you think?

### DIFF
--- a/buildcmd/nanospkg.go
+++ b/buildcmd/nanospkg.go
@@ -1,0 +1,200 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+
+	"github.com/nanovms/ops/cmd"
+	"github.com/nanovms/ops/lepton"
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	var rootCmd = &cobra.Command{Use: "nanospkg"}
+	rootCmd.AddCommand(BuildCommand())
+	rootCmd.Execute()
+}
+
+type PackageManifest struct {
+	Program string
+	Args    []string
+	Version string
+	Env     map[string]string
+}
+
+func BuildCommand() *cobra.Command {
+	var config string
+	var version string
+
+	var cmdBuild = &cobra.Command{
+		Use:   "build [binaryname] -c [configfile] -v [version]",
+		Short: "Build package from provided binary file",
+		Args:  cobra.MinimumNArgs(1),
+		Run:   runCmdBuild,
+	}
+
+	cmdBuild.PersistentFlags().StringVarP(&config, "config", "c", "", "ops config file")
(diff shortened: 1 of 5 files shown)


eyberg:
i think it's somewhat useful - as some binaries link to a lot of shit and I routinely screw up the paths 
when i'm building them /usr/lib vs /usr/local/lib

for example - w/tarantool ->

```
➜  ~  ops pkg contents tarantool_1.10.2 | grep File | wc -l
      53
```

, however, there are still a lot of things that this wouldn't automate - /etc files, libs that are assumed they are in the path but won't show up in ldd (eg: libnss), /usr/local/ config crap - i usually just need to deduce that from -d output so I don't know what the answer is to that -- this provides the 90% - and then you would use this as a jumpstart towards the last 10% of stuff you need to include?

as for where this (and other potential tools) like it might belong - i'm prob. in favor of having it live in this repo vs. a diff. repo so there's not so much code drift, but def. open to comments on this 

as for cli functionality - I'd prefer to keep the functionality limited in scope or at least as sub-commands - i don't want to overwhelm the user and there's a lot of intended functionality coming down the pipe as we roll in more orchestration support (the tool is called ops)

i guess to sum up - if it makes your life easier i'm ok w/including something like this - it's just more of a power user feature and should be positioned as such

eyberg:
also worth noting - we will sooner rather than later have a microsvc that deals w/user-supplied packages in a more sane manner so something to think about as this gets fleshed out

skupr:
Regarding scope of the tool
Creating a package usually consists of two phases:
1. Finding correct set of files & args & env to start progam correctly under nanos, and its usually done iteratively by `ops run -c config progname` with -d, as you mentioned.
2. After you have a working config & files, you need to place them properly, create manifest, readme and tar it up

This build cmd does not help phase 1 in any way but 99% automates phase 2.

There is only one thing that prob. needs to be done manually after build - remove arguments and files that were using for testing and should not go to public - like Hello.class for java, b.py for python etc.
But to avoid that one can develop a habit to put these additionals to -a argument and not in config.

Another, secondary thing to note - currently this build cmd populates package with generates and common files like libnss, certs, passwd - exactly same files that are added by `ops run`. And this creates some redundancy, as `ops load` will add them again. There's no problem with that for now however. We also might think of keeping this files generated in `ops run` and pkg build, but remove from `ops load` - this will make package execution process more transparent, imo.

eyberg:
going to close this for now as it's kinda old && was just a poc

PR #335, opened 2019-05-06
Don't chmod in linux getSharedLibs

heyitsanthony:
Calling os.Chmod will fail if the current user doesn't have write permission
for a given file, but can still run ldd (e.g., /bin/ls). This patch just
leaves the permissions alone and lets ldd fail if there's no execution
rights.

I see the chmod was added in 3a3ba2307b24aa15ed50d38d4800f5e9cd2ea76b but I can't figure out what path expects execute permissions to not be set. If a binary doesn't have +x set, it seems safer to have the user choose to grant the permission after seeing an error instead of automatically clobbering the mode with 755.

/cc @tijoytom

### DIFF
--- a/lepton/ldd_linux.go
+++ b/lepton/ldd_linux.go
@@ -37,19 +37,14 @@ func getSharedLibs(targetRoot string, path string) ([]string, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


eyberg:
thanks!

PR #349, opened 2019-05-22
quote/escape filenames in manifest if needed

### LINKED ISSUE #427: instances create order seems reversed/un-ordered
it's common to hit the deploy button a few hundred times in a day, by default creating an new instance should use the latest image - was pretty sure that the order here was created at desc (to by default always use latest) but not sure that's actually happening - need to look into

skupr:
This will allow to fix nanovms/nanos#427. Works only with nanovms/nanos#822

### DIFF
--- a/lepton/manifest.go
+++ b/lepton/manifest.go
@@ -211,7 +211,11 @@ func (m *Manifest) String() string {
 	sb.WriteString("arguments:[")
 	if len(m.args) > 0 {
 		fmt.Println(m.args)
-		sb.WriteString(strings.Join(m.args, " "))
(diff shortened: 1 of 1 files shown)


skupr:
We should probably also escape program arguments and environment variables - @mkhon what do you think?

eyberg:
+1 on the arguments - i noticed w/cassandra there was an argument ```-XX:OnOutOfMemoryError=kill -9 %p``` that was screwing things up

skupr:
@eyberg Added escaping of args and environment values in manifest generation

PR #353, opened 2019-06-05
Fix nanos go tests on macOS

mkhon:
Some files are added from $NANOS_TARGET_ROOT so check their existence also there.

AddDirectory() should also probably be fixed in this way but it never
was able to lookup in target root.


eyberg:
@mkhon ok to merge?

PR #360, opened 2019-06-18
Use dep for dependency management

ssudake21:


ssudake21:
Mod is enabled Go 11 and above. Most of the projects are moving to modules. Most of projects are moving to modules, we should surely move to modules in sometime. This would happen with migration to Golang 1.12 prob.

PR #367, opened 2019-06-25
Use "mkfs -b" to build the final image.

mkhon:
"mkfs -b" writes proper MBR which is required by some cloud providers (AWS).


mkhon:
It turns out proper MBR is not enough for AWS to import the image (it does more sanity checks on the image filesystems).
However "mkfs -b" eliminates the use of "cat" so I would merge this anyway.

PR #381, opened 2019-07-25
Add gdb enable and port flag to ops run

x64k:
Added support for enabling qemu's GDB interface. Only debugging over TCP
is supported (fortunately, it's the most common option, and the most
useful for us), and the user can specify the TCP port to be used for the
gdb interface. Specifying a port also enables the gdb interface --
specifying one without the other would be meaningless, since only TCP is
supported.

Signed-off-by: Alexandru Lazar <alazar@startmail.com>

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -37,6 +37,11 @@ func runCommandHandler(cmd *cobra.Command, args []string) {
 		panic(err)
 	}
 
+	gdbport, err := cmd.Flags().GetInt("gdbport")
+	if err != nil {
+		panic(err)
+	}
+
 	noTrace, err := cmd.Flags().GetStringArray("no-trace")
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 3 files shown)


x64k:
This came up while we were discussing https://github.com/nanovms/nanos/issues/367 . It's a reasonable band-aid until we can get gdb working in nanos.

I'm also thinking it would be useful to be able to specify qemu's -s flag, which prevents the CPU from actually running, so that you can connect via gdb and start the CPU yourself. However, that's a pretty dev-centric flag and I'm not sure if we want to explicitly expose that via ops. IIRC cobra supports "hidden" flags (i.e. parses them, but doesn't show them in the help output) but I'm not sure we really want that flag exposed in production *at all*.

x64k:
I'm a bit creeped out by the CI tests -- I saw this on my local machine as well:

`root_test.go:155: Get http://localhost:8080/example.html: read tcp 127.0.0.1:34648->127.0.0.1:8080: read: connection reset by peer`

I thought it was caused by my old, slow laptop and added some extra sleep before doing the HTTP request but now I see it on real machines, too. Is this a known problem, or am I doing something wrong here?

PR #486, opened 2020-04-09
Vultr delete image

bneil:
- added idea to .gitignore
- added ability to remove snapshot/bucket from vultr


bneil:
Not sure why the build failed, i generated the binary locally and it deletes fine

eyberg:
thanks - I think the test suite needs some work but this PR should be fine - want to close your other?

PR #487, opened 2020-04-10
feat(vultr): Added start/stop/delete server commands

bneil:


eyberg:
thanks! eventually we'll want to replace all this debug output under some sort of debug loglevel but this is fine for now

Issue #459, opened 2020-02-12
build / run / image create / load incongruencies


eyberg:
lots of flags are missing from build that are present in run

* run has a lot of stuff mainly cause it's the instant one-command experience to build/boot/run an image instantly

* build is basically the equivalent of run w/out the 'running' part

* load assumes pkgs are involved but has run like functionality where it assumes you are going to run it immediately rather than create an image

* create assumes you are uploading to the cloud

i don't think it's necessary to have all of this shit

PR #488, opened 2020-04-11
New command to allow starting/loading an instance solely from a config

mildmelon:
## Summary
I've created a new command `start` to enable loading an instance entirely from a config without the need for extra parameters, specifically the package name.

After scanning the source code for the `load` command there wasn't much purpose for specifically defining the package name every time, outside of semantics. For clarity and sharability, I wanted the ability to define the package name in the config itself.

## Usage
config.json
```
{
  "PackageName": "node_v11.5.0",
  "Args":["one","two"],
  "Dirs":["myapp/static"]
}
```
```
ops start -c config.json
```
Ideally
```
ops load -c config.json
```

## Reasoning
I've considered modifying the `load` command with the changes implemented here, but I wanted to show a new approach without modifying the existing context. I'd like to get feedback on your preference for whether these changes should be implemented as a new command or modifying the existing `load` command, (I will refactor my changes if they fit better into the `load` command).

I felt that the config should house all contextual information related to the VM. This change also helps shift the `load` command closer to how the `run` command works. Since it's currently not possible to build a bootable image for interpreted languages, I wanted something akin to a bootable image that was just as easy to use.

I think it's valuable to not worry about the base package used when loading an application (beyond the initial creation of the config) and would help ease any potential automated deployment issues and reduce the need to modify deployment code.

For example, `ops load -c config.json` is more resilient to language version changes than `ops load python_2.7.15rc1 -c config.json`.
Imagine you are working with Python 2, and have written deployment code that uses `ops load python_2.7.15rc1 -c config.json`. You use that command for various deploys: local development, staging server, production server. If a year or so down the line you want to migrate to Python 3, you'd need to explicitly update every place that uses that command. This runs the risk of inconsistent environments and potential production crashes. Now instead of explicitly defining the package name in various places, you have the package name in the config.json, checked into the application's VCS. This has the intended side effect of catching issues in local development before they even reach staging or production, and providing a consistent config across environments.

In summary, you can always take precedence over the package name parameter instead of the config package name, but giving the option to define the package name in a consistent place provides a lot of value during deployments.


eyberg:
a few thoughts:

1) this touches on my concerns expressed here: https://github.com/nanovms/ops/issues/459 - I feel there is definitely room for cleanup here - it is kind of the result of gradual evolution

2) I think it's perfectly fine to allow specifying the package in the config via ```ops load``` as you suggest rather than forcing it as a flag - will definitely accept that change if you want to do that - I think adding this functionality to ```ops build``` would be good to have too

3) "Since it's currently not possible to build a bootable image for interpreted languages, I wanted something akin to a bootable image that was just as easy to use."

can you elaborate on this? this is definitely achievable but I admit it's not super evident and could use some refactoring in how this works

all images wind up in ~/.ops/images/ regardless of what command is used and ```ops instance create``` can be used for interpreted languages/packages to build images to deploy to prod, if you use ```ops load``` to create an image from a package it has this crappy behavior where it creates an image with the name of the package - you can pass it ```-i``` though to give it a name

from there you can run ```ops run -s ``` to skip the re-build or 

```ops instance create -i myimage -t onprem``` to run your premade image (from the -i) locally

mildmelon:
Thank for getting back to me, regarding your thoughts:
1. I agree that there are some ambiguous and mismatching args, although I don't feel as though I have the knowledge to clean it up. 
2. Sounds great, I'll create a new branch to integrate the package name into the config, and create handlers for both `load` and `build`.
3. I was looking through the documentation, and it's a bit sparse regarding how images are created through the `load` command. My initial impression was that they would be built into the working directory, source: https://nanovms.gitbook.io/ops/faq#q-what-artifact-does-ops-produce

Yeah, it's mostly just that any docs on how to build the images using `load` don't exist (or I couldn't' find them). This might not be a priority, but you might want consider more evident and verbose docs regarding the nuances of the tool.

I think having a bit more documentation regarding around how the images are created and where they are stored would be valuable, ~~since coming from a Docker background I was expecting a way to see the built images (without needing to `ls ~/.ops/images`).~~ I now see that `ops image list -t onprem` lists local images.

Side note, while deleting onprem instances I noticed that they weren't being deleted even though I was passing the instance name, eg: `ops instance delete webserver -t onprem -z a`. This was a bit misleading, because after digging into the source, I found that the command actually takes the instances PID. My suggestion `ops instance delete <instance_name> [flags]` -> `ops instance delete <instance_pid> [flags]`.

eyberg:
yeh, thank you for noting the missing/incorrect docs, I agree, the docs, https://github.com/nanovms/ops-documentation could use some refreshing, we definitely accept PRs there too :)

as for deleting instances - the pid is unique - that's why it's being used locally versus the name, if you look at the aws && gcloud instance names aws uses tags and the gcloud instances tack on a timestamp (there are diff nuances here and also considerations that the dominant deploy type is public cloud not local) - all of this could use an overhaul w/something that's more common

eyberg:
this is super old so i'm going to close, we're still tracking this request though

PR #568, opened 2020-07-19
Disable spinner for pkg list

### LINKED ISSUE #250: disable spinner for pkg list
this is just a small annoyance - it's good to have the spinner for large images but 1.4k json file we porb. don't need this..

image

Regan-Koopmans:
These changes resolve #250 .

I have introduced a new parameter `showProgress` to the function `DownloadFile`. This allows the calling code to determine whether a progress spinner should print in the terminal for this download. I have added a useful helper function `DownloadFileWithProgress`, which conforms to the previous function signature, but sets `showProgress` to `true` automatically.

The instance where the package list should not display a spinner sets this explicitly to `false`. I can write a wrapper for this case too, but since this is the only occurrence it did not seem strictly essential at this stage. Let me know whether you would like me to add `DownloadFileWithoutProgress` as part of these changes too @eyberg ?

### DIFF
--- a/lepton/image.go
+++ b/lepton/image.go
@@ -108,7 +108,7 @@ func addDefaultFiles(m *Manifest, c *Config) error {
 
 	localtar := path.Join(GetOpsHome(), "common.tar.gz")
 	if _, err := os.Stat(localtar); os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


eyberg:
this is fine - thanks!

PR #592, opened 2020-08-28
add default config parser

### LINKED ISSUE #405: .opsrc
right now you can specify zone/project/cloud etc as flags or in config.json - i'd imagine the common case for most people is that these settings prob. don't  change that much and we should have a default in ~/.opsrc

ronaudinho:
Resolves #405 by adding default config reader which will read from, in order of priority:
1. config file define using flags
2. config file location set on `OPS_DEFAULT_CONFIG`
3. config file location default on `~/.opsrc`

hence if all 1, 2, and 3 exist, `ops` will use value from 1.

Currently needs feedback on the following:
- config file is in `json` format, but if i see `~/.opsrc`, i would assume it is in env file format, which does not support array and map (current default config is expected to be valid `json` as well)
- have not set configuration via env var (either do it via 1 or set it explicity), not sure about priority of these values as well (i.e if value is set, is it overridden by config file/CLI flags?)

On approval, will update https://github.com/nanovms/ops-documentation/issues/65 as well

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/user"
 	"path"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


eyberg:
yeh - in general config/cli should override - as for the env file format there's a handful of variables that would work there such as region but there's also an awful lot of vars that wouldn't work so i'm ok w/it being json (or if we really want to just explicitly rename it to .json to remove any confusion

PR #595, opened 2020-08-30
use kernel.img for nightly

### LINKED ISSUE #594: stage3.img does not exist in nightly
when using `nightly`, `stage3.img` is required but there is only `kernel.img` under `~/.ops/nightly`, not sure if this is the intended behavior

ronaudinho:
resolve #594

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -66,7 +66,7 @@ func TestDownloadImages(t *testing.T) {
 	// remove the files to force a download
 	// ignore any error from remove
 	boot := path.Join(api.GetOpsHome(), api.LatestReleaseVersion, "boot.img")
-	kernel := path.Join(api.GetOpsHome(), api.LatestReleaseVersion, "stage3.img")
+	kernel := path.Join(api.GetOpsHome(), api.LatestReleaseVersion, "kernel.img")
 	mkfs := path.Join(api.GetOpsHome(), api.LatestReleaseVersion, "mkfs")
 	os.Remove(mkfs)
 	os.Remove(boot)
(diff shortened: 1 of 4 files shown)


eyberg:
looks like it broke here -> https://github.com/nanovms/nanos/commit/f44463b1a2dfa9ad367ea4941cc5765690fecf3e - although versioned releases still use 'stage3'  - i'm ok w/this fix but i think it also makes sense to keep the name consistent so if @wjhun doesn't mind maybe you can fix the .circleci/config line there?

francescolavra:
The old stage3.img file name is not used anymore in the kernel, and the new kernel.img file name is now referenced in many places if the kernel Makefiles, so if we just change the .circleci/config line there, it would break the CI build.
So, either we revert all the file name changes in the kernel, or Ops will have to use the new file name.

eyberg:
that's fine - we can just update any name refs in this repo && cut a new release and it should be fine

eyberg:
@ronaudinho might be good just to rename all refs and then we'll merge this

```
➜  ops git:(master) ✗ grep -R stage3 * | grep -v vendor
cmd/root_test.go:       kernel := path.Join(api.GetOpsHome(), api.LatestReleaseVersion, "stage3.img")
cmd/utils.go:           c.Kernel = path.Join(api.GetOpsHome(), version, "stage3.img")
lepton/manifest_test.go:        kernel = `kernel:(contents:(host:stage3/stage3))
lepton/manifest_test.go:        m.AddKernel("stage3/stage3")
lepton/manifest_test.go:        m.AddKernel("stage3/stage3")
```

ronaudinho:
renamed all `stage3` and added a quick hack in the last commit, will take out if unnecessary.

eyberg:
i'm merging this now as I need to cut new releases for both nanos/ops

PR #598, opened 2020-09-03
kvm/qemu mounting

ronaudinho:
### RFC
1. currently using lousy implementation of `JSON` store to keep track of created volumes, maybe use some more established KV store?
2. `ops volume create` to create volume is necessary before running `ops run --mounts <uuid:mount_path>` since `UUID` is necessary. my idea is that when user wants to mount on run, they could perhaps just provide `host_path` and that new volume will be created and attached on the fly?

sample usage can be seen at https://github.com/ronaudinho/ops-examples/commit/722fff50a514c4f0e0693ee4280d66cc0c3a7002
(i tested `JSON` read/write, but on write, `json.Unmarshal` result is weird, so i do not call it in the main)

#### references
1. https://github.com/solo-io/unik/blob/master/pkg/client/volumes.go
2. https://github.com/solo-io/unik/blob/master/docs/cli.md#create-a-volume
3. https://github.com/firecracker-microvm/firecracker/pull/1351

### TODO
1. work with non-empty volume
2. update volume state on attach/detach
3. refactor volumeStore


lepton/qemu.go
```diff
@@ -440,6 +440,7 @@ func (q *qemu) addAccel() {
 func (q *qemu) setConfig(rconfig *RunConfig) {
 	// add virtio drive
 	q.addDrive("hd0", rconfig.Imagename, "none")
+	// q.addOption("-drive", fmt.Sprintf("if=none,id=hd0,format=raw,file=%s", rconfig.Imagename))
```
francescolavra, line 443: Did you forget to remove this commented out line?

lepton/volume.go
```diff
@@ -0,0 +1,427 @@
...
+		if err != nil {
+			return vol, err
+		}
+		if cur.ID == id {
+			vol = cur
+			continue
+		}
+		volumes = append(volumes, cur)
+	}
+	if reflect.DeepEqual(vol, NanosVolume{}) {
```
eyberg, line 385: not confident the reflection is necessary here - are you looking for an empty volume? you can just test for the presence of a variable instead

eyberg: if you need to ensure something is of the correct type just cast it ```newvar, ok := unknown.(newtype)```

lepton/qemu.go
ronaudinho: yes, fixed now

lepton/volume.go
ronaudinho: updated, separated store implementation to its own file

lepton/image.go
```diff
@@ -525,3 +556,19 @@ func lookupFile(targetRoot string, path string) (string, error) {
 
 	return path, err
 }
+
+// saveImageConfig saves image config as JSON
+// for volume attach/detach purposes
+func saveImageConfig(c Config) error {
```
eyberg, line 562: this is fine for now but something we should prob. fix in the future - a good chunk of the variables being stored here are dynamic in nature and don't belong here;

also - 'manifests' have an inherent meaning already and probably wise to change this to a different phrase

lepton/manifest.go
```diff
@@ -143,6 +152,62 @@ func (m *Manifest) AddDirectory(dir string) error {
...
 
+// AddRelativeDirectory adds all files in dir to image
+func (m *Manifest) AddRelativeDirectory(src string) error {
+	err := filepath.Walk(src, func(hostpath string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		var vmpath string
+		if hostpath[0] != '/' {
```
eyberg, line 163: dupe? not needed

eyberg (approved):
i'd dbl chk that dupe comment && might be worth adding a test of some kind but otherwise we can merge

eyberg:
merging since original comments were fixed && i've created tkts for the other issues and this has been sitting for a few weeks now

Issue #593, opened 2020-08-29
prefer accel over not


eyberg:
I think the default should be to have ops always running w/accel on and nag users that can't/won't instead

- [x] warn user if accel is set by ops but we can't actually use it (eg: no virt extensions because they are in a vm or no permission are set) see https://github.com/nanovms/ops/issues/440

- [x] if accel *can't* be set because it's in a vm it should tell the user to turn off accel and maybe explain why

- [x] pro-actively try to install ops https://github.com/nanovms/ops/blob/master/install.sh with the correct perms

as put into the attached pr I believe the reason why this has not been set as default is because the vast majority of users probably will not have these defaults set and will have a bad first time experience

emikhalev:
Is the issue resolved? Or is there are any additional work?

eyberg:
yes - thank you!

Issue #440, opened 2019-12-16
Wrong check for /dev/kvm permissions


nirs:
Using --accel fails on Fedora 30, complaining about user not in "kvm" group:

```
./ops run --accel -v hi
booting /home/nsoffer/.ops/images/hi.img ...
You specified hardware acceleration but you don't have rights.
Try adding yourself to the kvm group: `sudo adduser $user kvm`
You'll need to re login for this to take affect.
```

But /dev/kvm is accessible:
```
$ ls -lh /dev/kvm
crw-rw-rw-. 1 root kvm 10, 232 Dec 16 21:46 /dev/kvm
```

So the same command succeeds outside of ops:

```
$ $ qemu-system-x86_64 -drive file=/home/nsoffer/.ops/images/hi.img,format=raw,if=none,id=hd0 -device virtio-blk,drive=hd0 -device virtio-net,netdev=n0 -netdev user,id=n0 -nodefaults -no-reboot -device isa-debug-exit -m 2G -display none -serial stdio -enable-kvm -cpu host
assigned: 10.0.2.15
yo!
```

A better way to handle this would be to use:
````
-machine accel=kvm:tcg
````
So in the worst case there is a fallback to emulation, but the command does not fail.

Or use access() to check the permissions:

```
>>> os.access("/dev/kvm", os.R_OK | os.W_OK)
True
```

**Comment by eyberg** [comment]
thanks for the report, agreed this is a better method - didn't know about that fallback option

**Comment by jevgeniShupilov00** [comment]
On Ubuntu 20.04 vm-ware, I'm getting a general protection fault issue.
```
cpu: 0
 interrupt: 000000000000000d (General protection fault)
     frame: ffff800001460000
error code: 0000000000000000
```
Is it impossible to run ops on vmware? 
When runs qemu in outside of ops, it works for me

eyberg:
@jevgeniShupilov00 you won't be able to turn on kvm in vmware w/out nested virtualization turned on - tbh i have no clue if that is supported as it's not really something that we would support, this task kinda requires access to a real linux box

eyberg:
https://github.com/nanovms/ops/pull/629

PR #599, opened 2020-09-03
set accel by default

emikhalev:
This solves the issue https://github.com/nanovms/ops/issues/593

### DIFF
--- a/cmd/load.go
+++ b/cmd/load.go
@@ -238,7 +238,7 @@ func LoadCommand() *cobra.Command {
 	cmdLoadPackage.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "verbose")
 	cmdLoadPackage.PersistentFlags().BoolVarP(&bridged, "bridged", "b", false, "bridge networking")
 	cmdLoadPackage.PersistentFlags().StringVarP(&imageName, "imagename", "i", "", "image name")
-	cmdLoadPackage.PersistentFlags().BoolVarP(&accel, "accel", "x", true, "use cpu virtualization extension")
+	cmdLoadPackage.PersistentFlags().BoolVar(&accel, "accel", true, "use cpu virtualization extension")
 	cmdLoadPackage.PersistentFlags().BoolVarP(&skipbuild, "skipbuild", "s", false, "skip building package image")
 	cmdLoadPackage.PersistentFlags().BoolVarP(&local, "local", "l", false, "load local package")
 	return cmdLoadPackage
--- a/cmd/load.go
+++ b/cmd/load.go
@@ -238,7 +238,7 @@ func LoadCommand() *cobra.Command {
 	cmdLoadPackage.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "verbose")
 	cmdLoadPackage.PersistentFlags().BoolVarP(&bridged, "bridged", "b", false, "bridge networking")
 	cmdLoadPackage.PersistentFlags().StringVarP(&imageName, "imagename", "i", "", "image name")
-	cmdLoadPackage.PersistentFlags().BoolVarP(&accel, "accel", "x", false, "use cpu virtualization extension")
+	cmdLoadPackage.PersistentFlags().BoolVarP(&accel, "accel", "x", true, "use cpu virtualization extension")
 	cmdLoadPackage.PersistentFlags().BoolVarP(&skipbuild, "skipbuild", "s", false, "skip building package image")
 	cmdLoadPackage.PersistentFlags().BoolVarP(&local, "local", "l", false, "load local package")
 	return cmdLoadPackage
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -220,7 +220,7 @@ func RunCommand() *cobra.Command {
 	cmdRun.PersistentFlags().BoolVarP(&skipbuild, "skipbuild", "s", false, "skip building image")
 	cmdRun.PersistentFlags().StringVarP(&imageName, "imagename", "i", "", "image name")
 	cmdRun.PersistentFlags().StringVarP(&manifestName, "manifest-name", "m", "", "save manifest to file")
-	cmdRun.PersistentFlags().BoolVarP(&accel, "accel", "x", false, "use cpu virtualization extension")
+	cmdRun.PersistentFlags().BoolVarP(&accel, "accel", "x", true, "use cpu virtualization extension")
 	cmdRun.PersistentFlags().IntVarP(&smp, "smp", "", 1, "number of threads to use")
 
 	return cmdRun
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -220,7 +220,7 @@ func RunCommand() *cobra.Command {
 	cmdRun.PersistentFlags().BoolVarP(&skipbuild, "skipbuild", "s", false, "skip building image")
 	cmdRun.PersistentFlags().StringVarP(&imageName, "imagename", "i", "", "image name")
 	cmdRun.PersistentFlags().StringVarP(&manifestName, "manifest-name", "m", "", "save manifest to file")
-	cmdRun.PersistentFlags().BoolVarP(&accel, "accel", "x", true, "use cpu virtualization extension")
+	cmdRun.PersistentFlags().BoolVar(&accel, "accel", true, "use cpu virtualization extension")
 	cmdRun.PersistentFlags().IntVarP(&smp, "smp", "", 1, "number of threads to use")
 
 	return cmdRun
--- a/install.sh
+++ b/install.sh
@@ -205,7 +205,7 @@ ops_install_qemu() {
   fi
 
   if ! which qemu-system-x86_64>/dev/null; then
-    printf "QEMU not found. Please install qemu using your package manager and re-run this script"
+    printf "QEMU not found. Please install QEMU using your package manager and re-run this script"
   fi
 }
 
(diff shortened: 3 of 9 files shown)


eyberg (changes requested):
thanks! as innocuous as this issue sounds, part of the work here would be being 

1) warn user if accel is set but we can't (eg: no virt extensions cause they are in a vm, or more importantly, no permissions are set)

2) pro-actively try to install ops with accel avail (otherwise there is a good chance it won't work the first time)

I think the only reason this is the way it is is because we don't want users to have a bad first-time

emikhalev:
Hello Ian

Thank you for your comments. I would be grateful if you could clarify some details.

1. Regarding warning user if we can’t use HW acceleration. I’ve found only one place (for now) where we attempt to use acceleration: qemu.go:setAccel there are different types of errors (no support, no user permissions, or some internal errors). So the questions are:
	- Is it OK if I will add only warning only in qemu.go or you plan some refactoring to implement such check as general logic for all future hypervisors?
	- As there different reasons why we can’t use HW acceleration should we show only one message for all for or it will be better to give more detailed info about the reasons?

2. Regarding pro-actively try to install ops with HW acceleration support: do you meaning logic of update command?
If so, should we just update and ask the user to re-run command or (the better?) automatically re-run ops with the same args?

eyberg:
1) we have no plans to integrate with other hypervisors for local use, all the other hypervisors we support are considered remote (there is planned support for hyper-v but that's kind of a completely diff. issue and I wouldn't worry about that)

* I think it's is very important to give a precise error message on why someone can't have hardware acceleration otherwise this is going to cause more problems than it's worth, by default permissions aren't set and a lot of end-users will mistakenly believe that they can run this inside of a vm

2) no, I'm talking about modifying the install script https://github.com/nanovms/ops/blob/master/install.sh to do things like check if you have virt extensions available, if not warn the user that things will be slow && they have to use a flag to disable acceleration

emikhalev:
Hello Ian

I've added some changes. Please review them. Is this what you are thinking about?

eyberg (approved):

eyberg:
having some others chk it out, this might tie into https://github.com/nanovms/ops/issues/440

cmd/load.go
```diff
@@ -238,7 +238,7 @@ func LoadCommand() *cobra.Command {
 	cmdLoadPackage.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "verbose")
 	cmdLoadPackage.PersistentFlags().BoolVarP(&bridged, "bridged", "b", false, "bridge networking")
 	cmdLoadPackage.PersistentFlags().StringVarP(&imageName, "imagename", "i", "", "image name")
-	cmdLoadPackage.PersistentFlags().BoolVarP(&accel, "accel", "x", false, "use cpu virtualization extension")
+	cmdLoadPackage.PersistentFlags().BoolVarP(&accel, "accel", "x", true, "use cpu virtualization extension")
```
francescolavra, line 241: As far as I can see, the shorthand version of a boolean flag with true as default value doesn't make much sense, because the shorthand version can only be used to set the flag to true. So I would drop the shorthand flag "x" and use BoolVar() instead.

cmd/run.go
```diff
@@ -220,7 +220,7 @@ func RunCommand() *cobra.Command {
 	cmdRun.PersistentFlags().BoolVarP(&skipbuild, "skipbuild", "s", false, "skip building image")
 	cmdRun.PersistentFlags().StringVarP(&imageName, "imagename", "i", "", "image name")
 	cmdRun.PersistentFlags().StringVarP(&manifestName, "manifest-name", "m", "", "save manifest to file")
-	cmdRun.PersistentFlags().BoolVarP(&accel, "accel", "x", false, "use cpu virtualization extension")
+	cmdRun.PersistentFlags().BoolVarP(&accel, "accel", "x", true, "use cpu virtualization extension")
```
francescolavra, line 223: Same as above.

install.sh
```diff
@@ -297,6 +297,49 @@ ops_link() {
...
+checkHWAccelSupport() {
+  echo ""
+
+  local hwSupported=false
+  local haveRights=false
+
+  if [ "$OS" = "linux" ]; then
+    local group="kvm"
+    local username=`whoami`
+    local hw_support="grep -woE 'svm|vmx' /proc/cpuinfo | uniq"
```
francescolavra, line 309: The shell command must be enclosed in ``, not "".

install.sh
```diff
@@ -297,6 +297,49 @@ ops_link() {
...
+
+  if [ "$hwSupported" = "false" ]; then
+        echo "$yellow> Hardware acceleration not supported by your system. $reset"
+        echo "$yellow> Ops will attempt to enable acceleration by default and will show a warning if the system doesn't support it. $reset"
+        echo "$yellow> To avoid such warnings you may disable acceleration in configuration or via command line parameters. $reset"
+  else
+    if [ "$haveRights" = "false" ]; then
+      echo "$yellow> Hardware acceleration is supported, but you don't have rights. $reset"
+      echo "$yellow> Try adding yourself to the kvm group: sudo adduser ${username} kvm $reset"
+      echo "$yellow> You'll need to re-login for this to take affect. $reset"
```
francescolavra, line 338: Replace "affect" with "effect" (I know this spelling error was there before, but here's an opportunity to correct it).

lepton/qemu_errors.go
```diff
@@ -0,0 +1,62 @@
...
+		targetErrQemuCannotExecute           *errQemuCannotExecute
+		targetQemuHWAccelNoUserRights        *errQemuHWAccelNoUserRights
+		targetQemuHWAccelNotSupported        *errQemuHWAccelNotSupported
+	)
+	if errors.As(err, &targetErrQemuHWAccelDisabledInConfig) {
+		return fmt.Sprintf(WarningColor, "You have disabled hardware acceleration\n"), false
+	}
+	if errors.As(err, &targetErrQemuNotInstalled) {
+		return fmt.Sprintf(WarningColor, "Cannot find QEMU (looks like it is not installed)\n"+
+			"Please install qemu using your package manager and re-run current command\n"), true
```
francescolavra, line 44: Replace "qemu" with "QEMU", for consistency with the previous line.

lepton/qemu_errors.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	if errors.As(err, &targetErrQemuCannotExecute) {
+		return fmt.Sprintf(WarningColor, "QEMU installed, but cannot be executed\n"+
+			"Please check current user rights\n"), true
+	}
+
+	if errors.As(err, &targetQemuHWAccelNoUserRights) {
+		return fmt.Sprintf(WarningColor, "You specified hardware acceleration, but you don't have rights\n"+
+			"Try adding yourself to the kvm group: `sudo adduser $user kvm`\n"+
+			"You'll need to re-login for this to take affect\n"), true
```
francescolavra, line 54: Now, the statement "You specified hardware acceleration" could be wrong, since accel is enabled by default. So I would use something like "You don't have rights for using hardware acceleration\n" instead, and I would not consider this warning to be fatal (i.e., replace `true` with `false`).
Also, replace "affect" with "effect".

lepton/qemu_errors.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	if errors.As(err, &targetQemuHWAccelNotSupported) {
+		return fmt.Sprintf(WarningColor, "You specified hardware acceleration, but it is not supported\n"), false
```
francescolavra, line 58: Also here, the statement "You specified hardware acceleration" could be wrong. I would use something like  		"Hardware acceleration is not supported\n".

lepton/qemu_errors.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	return fmt.Sprintf(WarningColor, "Harware acceleration cannot be used on the current host"), false
```
francescolavra, line 61: "Harware" -> "Hardware"

lepton/qemu.go
```diff
@@ -397,44 +401,55 @@ func (q *qemu) addOption(flag, value string) {
 	q.flags = append(q.flags, option)
 }
 
-func (q *qemu) addAccel() {
+func (q *qemu) setAccel(rconfig *RunConfig) error {
+	if rconfig.Accel || rconfig.Bridged {
+		return q.addAccel()
```
francescolavra, line 406: I don't know why acceleration is being silently enabled in a bridged configuration, but now that the accel flag is true by default, if the flag is false it means the user explicitly disabled acceleration in the config, so I think we should not enable it if a bridged config is there. So I would drop the `rconfig.Bridged` part in the `if` condition.

cmd/load.go
emikhalev: Agree, I've changed to BoolVar

cmd/run.go
emikhalev: The same

install.sh
emikhalev: Thanks, changed

install.sh
emikhalev: done

lepton/qemu_errors.go
emikhalev: done

lepton/qemu_errors.go
emikhalev: done

lepton/qemu_errors.go
emikhalev: done

lepton/qemu_errors.go
emikhalev: thank you. fixed

lepton/qemu.go
emikhalev: agree, done

eyberg:
thanks!

Issue #580, opened 2020-08-12
add support for GOVC_USERNAME && GOVC_PASSWORD if not set in GOVC_URL


eyberg:
https://github.com/nanovms/ops/issues/578

amedisetti:
Thanks for working on this fix , please let me know when would it be available.

PR #600, opened 2020-09-03
add support for GOVC_USERNAME and GOVC_PASSWORD

### LINKED ISSUE #578: Vsphere target is failing with datacenter
I set GOVC_URL, GOVC_DATACENTER,
When i run : ops image create -c config.json -t vsphere -a gtest, it is failing with:
Please specify datacenter 

I am following example given in doc.

ronaudinho:
resolves #578 and #580

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -90,30 +90,30 @@ func setDefaultImageName(cmd *cobra.Command, c *api.Config) {
 }
 
 // TODO : use factory or DI
-func getCloudProvider(providerName string) api.Provider {
+func getCloudProvider(providerName string) (api.Provider, error) {
 	var provider api.Provider
 
-	if providerName == "gcp" {
+	switch providerName {
+	case "gcp":
 		provider = &api.GCloud{}
-	} else if providerName == "onprem" {
+	case "onprem":
 		provider = &api.OnPrem{}
-	} else if providerName == "aws" {
+	case "aws":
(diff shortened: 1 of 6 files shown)


cmd/utils.go
```diff
@@ -93,21 +93,22 @@ func setDefaultImageName(cmd *cobra.Command, c *api.Config) {
 func getCloudProvider(providerName string) api.Provider {
 	var provider api.Provider
 
-	if providerName == "gcp" {
+	switch providerName {
```
ronaudinho, line 96: `switch` looks cleaner, if you don't mind the change

lepton/vsphere.go
```diff
@@ -116,11 +132,6 @@ func (v *Vsphere) Initialize() error {
 		This: *v.client.ServiceContent.SessionManager,
 	}
 
-	req.UserName = u.User.Username()
```
ronaudinho, line 119: `u.User` is set to nil in `soap.NewClient`

eyberg (changes requested):
I don't think this is going to work; right now we only support GOVC_URL with url-encoded login/pass inside the url; what needs to happen is if the l/p is not set in the url it should fall back to GOVC_USERNAME && GOVC_PASSWORD (which for some quirky reason) doesn't need to be url-encoded (prob. cause of the cast) - for more ref: https://github.com/vmware/govmomi/tree/master/govc

PR #612, opened 2020-09-22
add a new method to get all instances on Google Cloud Platform

minhkhiemm:


lepton/gcp.go
```diff
@@ -450,6 +450,40 @@ func (p *GCloud) ListInstances(ctx *Context) error {
 	return nil
 }
 
+// GetInstances return all instances on GCloud
+func (p *GCloud) GetInstances(ctx *Context) ([]compute.Instance, error) {
```
eyberg, line 454: this needs to be called from ListInstances and is missing quite a few things like pub/priv ip, created_at date, status, etc..

that could/should be a common struct that is re-used amongst all the various providers

eyberg (changes requested):

Issue #611, opened 2020-09-21
volume list can use the tablewriter instead of tabs


eyberg:
for example https://github.com/nanovms/ops/blob/master/lepton/gcp.go#L238

PR #618, opened 2020-09-22
Use tablewriter as volume list (#611)

u5surf:
Fixes: #611 

I replaced tab-fmt.Print style with tablewriter in volume `GetAll`.

### DIFF
--- a/lepton/volume.go
+++ b/lepton/volume.go
@@ -2,7 +2,6 @@ package lepton
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
@@ -11,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/go-errors/errors"
+	"github.com/olekukonko/tablewriter"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


lepton/volume.go
```diff
@@ -127,9 +127,14 @@ func (v *Volume) GetAll() error {
...
+	table.SetHeader([]string{"NAME", "UUID", "PATH"})
 	for _, vol := range vols {
-		fmt.Printf("%s\t\t\t%s\t\t\t%s\n", vol.Name, vol.ID, vol.Path)
+		var row []string
+		row = append(row, vol.Name)
+		row = append(row, vol.ID)
+		row = append(row, vol.Path)
+		table.Append(row)
 	}
 	return nil
```
eyberg, line 139: did you try this out? looks like it's missing a table.Render...

eyberg (changes requested):

lepton/volume.go
u5surf: @eyberg 
Sorry, it was missing to git add. But I added it right now.

eyberg (approved):

eyberg:
thanks!

PR #623, opened 2020-09-23
Print volume size in volume list table

### LINKED ISSUE #608: add size column to volume list
there are human readable functions in helpers.go

### LINKED ISSUE #638: Volume size is set to empty when not passing a size to the create command
## Description
When no volume size is passed to the `volume create` the volume size is set to `""` in the  `volumes.json` file.

## Steps to reproduce

1. create a volume called `test` without specifying the size of it
```bash
$ ./ops volume create test
```

2. Inspect the value of its size in the `volumes.json` file.
```bash
$ cat ~/.ops/volumes/volumes.json
```
It will be  `""`. This is a mistake since the minimum size of a volume in mkfs is 1 MB. 
```bash
~/.ops/0.1.27/mkfs -e empty.raw
```

ahmedaabouzied:
## Description
Prints the volume size alongside other values in the list volumes table. It first parses the size value into a human readable format using the helper functions in `lepton/helpers.go`.
Example
```bash
./ops volume list
+----------+--------------------------------------+--------+------------------------------------------------+
|   NAME   |                 UUID                 |  SIZE  |                      PATH                      |
+----------+--------------------------------------+--------+------------------------------------------------+
| test_vol | 9aa1a2a1-4d6d-a340-03d2-a77b05484c16 | 1.0 kB | /home/ahmedaabouzied/.ops/volumes/test_vol.raw |
+----------+--------------------------------------+--------+------------------------------------------------+
```
If there is an error parsing the size value of a certain volume, it prints an error instead indicating the name, ID of the volume with the invalid value as follows
```bash
./ops volume list
2020/09/23 04:45:57 invalid size value for volume test_vol with UUID 1352848a-547a-c67c-fce9-5f1e8e52bc8e: strconv.ParseFloat: parsing "": invalid syntax
```
Fixes #608 

### UPDATE
Fixes #638 as well

### DIFF
--- a/cmd/volume.go
+++ b/cmd/volume.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"log"
 	"path"
+	"strconv"
 
 	api "github.com/nanovms/ops/lepton"
 	"github.com/spf13/cobra"
@@ -52,7 +53,7 @@ func volumeCreateCommand() *cobra.Command {
 		Args:  cobra.MinimumNArgs(1),
 	}
 	cmdVolumeCreate.PersistentFlags().StringVarP(&data, "data", "d", "", "volume data source")
-	cmdVolumeCreate.PersistentFlags().StringVarP(&size, "size", "s", "", "volume initial size")
+	cmdVolumeCreate.PersistentFlags().StringVarP(&size, "size", "s", strconv.Itoa(api.MinimumVolumeSize), "volume initial size")
 	return cmdVolumeCreate
 }
 
--- a/lepton/volume.go
+++ b/lepton/volume.go
@@ -128,11 +128,12 @@ func (v *Volume) GetAll() error {
 		return err
 	}
 	table := tablewriter.NewWriter(os.Stdout)
-	table.SetHeader([]string{"NAME", "UUID", "PATH"})
+	table.SetHeader([]string{"NAME", "UUID", "SIZE", "PATH"})
 	for _, vol := range vols {
 		var row []string
 		row = append(row, vol.Name)
 		row = append(row, vol.ID)
+		row = append(row, v.parseSize(vol))
 		row = append(row, vol.Path)
 		table.Append(row)
 	}
@@ -252,4 +253,16 @@ func (v *Volume) AttachOnRun(mount string) error {
 	conf.Mounts[uuid] = path
 	conf.RunConfig.Mounts = append(conf.RunConfig.Mounts, vol.Path)
 	return nil
+}
+
+func (v *Volume) parseSize(vol NanosVolume) string {
+	if vol.Size == "" {
+		return bytes2Human(MiByte)
+	}
+	bytes, err := parseBytes(vol.Size)
+	if err != nil {
+		log.Printf("warning: invalid size value for volume %s with UUID %s: %s", vol.Name, vol.ID, err.Error())
+	}
+	size := bytes2Human(bytes)
+	return size
 }
--- a/lepton/volume.go
+++ b/lepton/volume.go
@@ -18,6 +18,9 @@ var (
 	localVolumeData = path.Join(GetOpsHome(), "volumes", "volumes.json")
 )
 
+// MinimumVolumeSize is the minimum size of a volume created with mkfs (1 MB).
+const MinimumVolumeSize = MByte
+
 var (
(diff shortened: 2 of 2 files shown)


eyberg:
so this is kind of separate from this pr/tkt but a volume is always at least 1mb in size

```
~/.ops/0.1.27/mkfs -e empty.raw
```

so the method of storing the volume info in ~/.ops/volumes is not quite correct - there should never be an empty, and even if there was this currently stop the listing of the other volumes w/an error which it should prob. do best effort and then list the error,

i'd just make it a requirement that it is of a correct size before even creating

ahmedaabouzied:
I see your point. This implementation returns an error if only one volume size value is corrupted. Ideally, we should print all the right modules and only print warnings about the corrupted modules, right?

PS Sorry I took this long to respond. I was feeling a bit sick for the past couple of days.

ahmedaabouzied:
I can work on making sure the volume size gets stored correctly in `volumes.json` in a separate issue. But we should handle corrupted volume size values here in the list method as well. Because even after making sure the right value gets saved by the `create ` volume command, a user can always change the fields in the `volumes.json` file manually. So this case should be handled. 

I believe also we can write some unit tests in for this method. I'll work on that too.

ronaudinho:
@ahmedaabouzied right now size is only stored if it is supplied in the flags on `ops volume create`, so a quick workaround that i can think of to ensure volume size is stored properly is to do a `os.Stat` to get the size after successfully creating volume (i am not sure if `dump` tool from `nanos` can do this)

> Because even after making sure the right value gets saved by the create volume command, a user can always change the fields in the volumes.json file manually

this storage solution is a placeholder an open for better implementation (i was thinking maybe we use existing KV libraries, but prefer not to add dependencies for now, since I would guess that most cloud-related volume functionalities will not need this unless we are storing the data locally as well)

> I believe also we can write some unit tests for this method. I'll work on that too.
i changed quite a bit of `lepton/volume.go` in #636 so probably would be better if you wait until it is merged

ahmedaabouzied:
@ronaudinho, yea I think we should wait after #636  is merged then we can add unit tests. For now, I believe we can just print a warning for if the size is incorrect

eyberg:
* there should never be an empty - the smallest size is 1mb - so if not specified that is the default size
* user's shouldn't manually editing that file or whatever it turns into so I wouldn't worry about that

ahmedaabouzied:
something like this maybe ?
```bash
2020/09/25 17:20:56 warning: invalid size value for volume test_vol with UUID 9aa1a2a1-4d6d-a340-03d2-a77b05484c16: strconv.ParseFloat: parsing "": invalid syntax
+----------+--------------------------------------+------+------------------------------------------------+
|   NAME   |                 UUID                 | SIZE |                      PATH                      |
+----------+--------------------------------------+------+------------------------------------------------+
| test_vol | 9aa1a2a1-4d6d-a340-03d2-a77b05484c16 | 0 B  | /home/ahmedaabouzied/.ops/volumes/test_vol.raw |
+----------+--------------------------------------+------+------------------------------------------------+
```

ahmedaabouzied:
Ok so it should print `1mb` if the size is not specified .. got it

ahmedaabouzied:
Now it prints the default size if not specified in the config. If all is well, I'll add some comments in the code to explain this and squash those two commits to for cleaner git history.

eyberg:
I don't think this type of error handling really belongs here - the real fix is to assign a default size to volumes of 1mb and make it to where you can't even create a volume w/out that

ahmedaabouzied:
@eyberg, I agree. But I'll open a separate issue for that to keep things organized.

ahmedaabouzied:
In the last commit, I made sure that the command `create` has a default value for the `-s` flag. If no value was passed, it assigns it to a string value equivalent to 1MB. 

I also made sure that it's not a static `1MB` value as this will cause tight coupling. It's a constant called `MinimumVolumeSize` exported by the lepton package from the volume.go file. And this constant just uses the definition of 1MB from the helpers.go file.

I think this solves the problem to some extent, right @eyberg , @ronaudinho ?

ahmedaabouzied:
Let me know if we need to change these 2 commit messages. I don't know the commit message rules followed here.

eyberg:
this is fine for now but we are going to roll out some new nanos changes here shortly on volumes so I wouldn't spend too much time here until that arrives

PR #629, opened 2020-09-24
check user have permissions for /dev/kvm

emikhalev:
check for /dev/kvm user permissions instead of the check is a user in KVM group, try to activate HW accel anyway: -machine accel=kvm:tcg (#440)

### DIFF
--- a/lepton/qemu.go
+++ b/lepton/qemu.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
-	"os/user"
 	"path"
 	"path/filepath"
 	"regexp"
@@ -16,6 +15,8 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+
+	"golang.org/x/sys/unix"
 )
 
 const qemuBaseCommand = "qemu-system-x86_64"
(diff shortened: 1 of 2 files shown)


emikhalev:
I've updated the code for checking KVM access, also trying to activate HW accel anyway. I'm not sure if we need to add any changes to the install.sh script (I think that checking for the KVM group is good enough). 

I've checked on the Ubuntu 18.04.5 LTS (GNU/Linux 4.15.0-117-generic x86_64) and it will be helpful if somebody can check on any other OS

lepton/qemu.go
```diff
@@ -306,39 +306,91 @@ func parseQemuVersion(data []byte) string {
...
+	}
+
+	filePerm := fi.Mode().Perm()
+	ownerPerm := (filePerm & 0700) >> 6
+	groupPerm := (filePerm & 0070) >> 3
+	othersPerm := filePerm & 0007
+
+	// Now we need file stat to continue
+	fsys := fi.Sys()
+	fStatT, ok := fsys.(*syscall.Stat_t)
```
eyberg, line 348: this by in large looks great - the only comment I have is - isn't there a native go stat call you could use here instead of this?

eyberg (approved):

francescolavra:
I checked with Ubuntu 16.04 LTS, and it works. However, I think the code could be simplified by using the access() syscall to check for /dev/kvm read/write permissions, as in the Go equivalent of `access("/dev/kvm", R_OK | W_OK)` (there are cases where a user does not appear to be in the group of a file but still has access to the file, e.g. with the setup below I'm able to use KVM even though the `root` group doesn't show up in my account's groups (there is the `sudo` group instead):
```
francesco@ubuntu:~/.ops/instances/java$ ll /dev/kvm
crw-rw----+ 1 root root 10, 232 Sep 16 07:29 /dev/kvm
francesco@ubuntu:~/.ops/instances/java$ groups
francesco adm dialout cdrom sudo dip plugdev lpadmin scanner saned sambashare kvm wireshark pgrimaging
```
But thanks to the fact that now ops tries to enable acceleration anyway, I can still run a VM with acceleration.

lepton/qemu.go
emikhalev: I've made changes according to @francescolavra comment

emikhalev:
> I checked with Ubuntu 16.04 LTS, and it works. However, I think the code could be simplified by using the access() syscall to check for /dev/kvm read/write permissions, as in the Go equivalent of `access("/dev/kvm", R_OK | W_OK)`

Thank you. I've made some changes, please review it. Sorry, I don't have very much experience with syscall and *nix API.

francescolavra:
Looks good to me, thanks.

Issue #499, opened 2020-04-28
ops should error out nicely for shared libs not found


eyberg:
ops currently gives a misleading error message when it can't resolve a shared lib

```
couldn't stat file not: No such file or directory

*exec.ExitError exit status 2
/home/eyberg/go/src/github.com/nanovms/ops/lepton/image.go:30 (0x1025f7f)
        BuildImage: if err = buildImage(&c, m); err != nil {
/home/eyberg/go/src/github.com/nanovms/ops/cmd/run.go:16 (0x10e24b8)
        buildImages: err := api.BuildImage(*c)
/home/eyberg/go/src/github.com/nanovms/ops/cmd/run.go:141 (0x10e3436)
        runCommandHandler: buildImages(c)
/home/eyberg/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:830 (0x106581d)
        (*Command).execute: c.Run(c, argWoFlags)
/home/eyberg/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 (0x106641b)
        (*Command).ExecuteC: err = cmd.execute(flags)
/home/eyberg/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864 (0x10e6a47)
        main: _, err := c.ExecuteC()
/home/eyberg/go/src/github.com/nanovms/ops/ops.go:8 (0x10e6a3e)
        main: cmd.GetRootCommand().Execute()
/usr/local/go/src/runtime/proc.go:203 (0x437382)
        main: fn()
/usr/local/go/src/runtime/asm_amd64.s:1373 (0x464c61)
        goexit: BYTE    $0x90   // NOP
```

```
eyberg@s1:~/dz$ ldd bob
        linux-vdso.so.1 (0x00007fffeb9ab000)
        libphobos2-ldc-shared.so.78 => not found
        libdruntime-ldc-shared.so.78 => not found
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2b6fd8f000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f2b7019d000)
```

in this case it should complain nicely mentioning what library it can't find

eyberg:
to repro:

```
eyberg@box:~/badlink$ cat main.c
#include <stdio.h>
#include "missing.h"

int main() {
  missing();
  printf("test");
}
eyberg@box:~/badlink$ cat missing.c
#include <stdio.h>

void missing(void)
{
[...]

PR #630, opened 2020-09-24
Fix/library

EduardGaifutdinov:


eyberg (changes requested):
needs to work for example pasted in #499

Issue #585, opened 2020-08-20
gcloud multiple volume mounting


eyberg:

eyberg:
https://github.com/nanovms/nanos/pull/1303

PR #636, opened 2020-09-25
GCP mounting

ronaudinho:
#585 
1. add `create`, `list`, `delete`, `attach`, and `detach` volume functionalities for GCP
2. additionally, refactor `GCloud` functions to reuse `compute.Service`

changes to non gcp-related providers are merely stubs to make `Provider` implements volume-related functionalities.

### DIFF
--- a/lepton/gcp.go
+++ b/lepton/gcp.go
@@ -15,7 +15,8 @@ import (
 	"github.com/olekukonko/tablewriter"
 
 	"golang.org/x/oauth2/google"
-	compute "google.golang.org/api/compute/v1"
+	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/option"
 )
 
 // GCloudOperation status check
@@ -71,6 +72,7 @@ func (gop *GCloudOperation) isDone(ctx context.Context) (bool, error) {
 // GCloud contains all operations for GCP
 type GCloud struct {
 	Storage *GCPStorage
+	compute *compute.Service
 }
 
 func (p *GCloud) getArchiveName(ctx *Context) string {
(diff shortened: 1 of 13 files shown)


lepton/volume.go
```diff
@@ -32,10 +31,19 @@ var (
 	errInvalidMountConfiguration = func(mount string) error { return errors.Errorf("%s: invalid mount configuration", mount) }
 )
 
-// Volume is service for managing nanos-managed volumes
-type Volume struct {
-	config *Config
-	store  volumeStore
+// VolumeService interface for volume related operations
+type VolumeService interface {
```
ronaudinho, line 35: since this is used by `Provider`, maybe move this to `lepton/provider.go`

lepton/gcp.go
```diff
@@ -681,3 +586,177 @@ func createArchive(archive string, files []string) error {
 func (p *GCloud) ResizeImage(ctx *Context, imagename string, hbytes string) error {
 	return fmt.Errorf("Operation not supported")
 }
+
+// CreateVolume creates local volume and converts it to GCP format before orchestrating the necessary upload procedures
+// TODO treat GCS object and Compute Image as temporary and delete after volume is created?
```
eyberg, line 591: yes, please delete the tmp. disk image and use the volume as the new artifact - if one wants to clone it or version it or whatever they should use the volume as the end artifact to do so

lepton/gcp.go
```diff
@@ -681,3 +586,177 @@ func createArchive(archive string, files []string) error {
...
+	vol = NanosVolume{
+		ID:   strconv.Itoa(int(d.Id)),
+		Name: d.Name,
+		Path: d.SelfLink,
+	}
+	return vol, nil
+}
+
+// DeleteVolume deletes specific volume in GCP
+// TODO delete GCS object and Compute Image as well? - related to CreateVolume
```
eyberg, line 705: yes

eyberg:
not sure where this is happening but need a better error msg if the projectId is not set for gcloud

for instance with volume list we get an error that tells us nothing but with instance list it explicitly states projectid is missing

```
➜  g2 ops volume list -t gcp
2020/09/25 10:05:10 googleapi: Error 400: Invalid resource field value in the request., invalidParameter

➜  g2 ops instance list -t gcp
projectid argument missing
```

eyberg:
so this doesn't work as-is - to make it work under current conditions you'd need modify the base image with the new mount point and re-deploy, but i really don't like that idea at all, so let's pause on this for now, and we'll refactor nanos to accomodate for attaching them without modifying the base image

eyberg (changes requested):
pause until nanos is refactored to allow volumes to attach w/out re-building underlying base image

lepton/gcp.go
ronaudinho: alright will update this

ronaudinho:
closing since #662 makes this obsolete

Issue #517, opened 2020-05-13
image extensions should be standardized


eyberg:
right now we don't really enforce naming conventions even though all the images are standard disk images - it's prob. worth specifying them as .img or something and I believe currently ```ops image list -t onprem``` and the equivalent run only finds them if they are comosed w/ '.img' extensions

right now on gcloud ->

```
➜  ~  ops image list
+-----------------------+--------+-------------------------------+
|         NAME          | STATUS |            CREATED            |
+-----------------------+--------+-------------------------------+
| bob2                  | READY  | 2020-04-07T10:40:30.004-07:00 |
+-----------------------+--------+-------------------------------+
| bobrandom-image       | READY  | 2020-02-26T09:33:57.598-08:00 |
+-----------------------+--------+-------------------------------+
```

and instances are brought up w/timestamps slapped on them to diff.

and aws has the same done for images

```
➜  ~  ops image list -t aws -z us-west-1
+------------+-------------------------------+-----------+--------------------------+
|    NAME    |              ID               |  STATUS   |         CREATED          |
+------------+-------------------------------+-----------+--------------------------+
| node-image | node-image1588609143075943000 | available | 2020-05-04T16:19:03.000Z |
```

so I think for local we should just keep the .img extension but list w/out the extension and same for cloud deploys

eyberg:
the other thing is that for both (incoming) esx && hyper-v support those will have diff. formats and might? store diff. formats locally

eyberg:
https://github.com/nanovms/ops/issues/1336 - seems like on-prem images are the odd ones out

PR #645, opened 2020-09-27
add image sync command

### LINKED ISSUE #502: ops image create should be able to consume from ~/.ops/images rather than forcing a pkg or the elf in cwd
this has been brought up a few times

right now there is an assumption that images are built on-demand - this works for quite a few purposes - you either pass it a package name or your binary, however there is a need in some cases to copy the image around and then deploy it w/out having access to the original binary/pkg/build env

packages aren't actually disk images - they are templates to build disk images from

```ops load``` && ```ops image``` create disk images on the fly

support for running pre-made images exists when running locally via:

```
ops instance create -i myimage -t onprem
```

but we are currently missing support for public cloud deploys

what is needed is a way for

```
ops image create
```

to be passed a locally stored image (found in ~/.ops/images) and use that instead of the mandatory pkg/binary for public cloud deploys

cc @mildmelon && @Dmitry-N-Medvedev

### PR DESCRIPTION
**Pull request by ronaudinho**
resolves #502 

`ops image sync <image_name> -t <cloud_provider>` allows user to transfer locally existing image to target cloud provider

since original `ops image` does not accept flags for some provider-related configs, user will need to set said fields in config

### DIFF
--- a/lepton/onprem.go
+++ b/lepton/onprem.go
@@ -113,6 +113,31 @@ func (p *OnPrem) DeleteImage(ctx *Context, imagename string) error {
 	return nil
 }
 
+// SyncImage syncs image from onprem to target provider provided in Context
+func (p *OnPrem) SyncImage(config *Config, target Provider, image string) error {
+	imagePath := path.Join(localImageDir, image+".img")
+	_, err := os.Stat(imagePath)
+	if err != nil {
+		return nil
+	}
+	config.RunConfig.Imagename = imagePath
+	config.CloudConfig.ImageName = image
+
+	// customizes image for target
+	ctx := NewContext(config, &target)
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: lepton/onprem.go [hunk 1]
```diff
@@ -113,6 +113,31 @@ func (p *OnPrem) DeleteImage(ctx *Context, imagename string) error {
 	return nil
 }
 
+// SyncImage syncs image from onprem to target provider provided in Context
+func (p *OnPrem) SyncImage(config *Config, target Provider, image string) error {
+	imagePath := path.Join(localImageDir, image+".img")
```
ronaudinho, line 118: when running `ops image sync <image_name>`, omit `.img`
might need change, related to #517

eyberg (approved):

PR #646, opened 2020-09-28
Return empty list if volumes config file not found

### LINKED ISSUE #640: running volumes list w/out creating volume shows unfriendly error msg
should just show empty list instead

```
eyberg@box:~$ ops volume list
2020/09/25 18:36:31 open /home/eyberg/.ops/volumes/volumes.json: no such file or directory
```

ahmedaabouzied:
## Description
This gets rid of the `no such file or directory error` that is printed if the `volumes.json` file doesn't exist. If this filed doesn't exist this will print an empty list in the volumes list table.

This uses `errors.Is()` instead of comparing errors by the `==` operator as this recommended [here](https://golang.org/pkg/errors/#pkg-overview). 

Fixes #640

### DIFF
--- a/lepton/store.go
+++ b/lepton/store.go
@@ -3,6 +3,7 @@ package lepton
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


eyberg:
thanks! there's a lot of crappy non-idiomatic error handling in the codebase that needs to be refactored out, also as a heads up - there's going to be new volume work landing in nanos hopefully this wk so might want to wait until that happens before doing much more volume work

PR #649, opened 2020-09-28
Fix/ipv6

EduardGaifutdinov:


eyberg:
this unilaterally stops any ipv6 from working - only user-mode is the problem (eg: any cloud provider should work just fine) - further this is on a bridge which explicitly doesn't use user-mode..

PR #661, opened 2020-10-02
implements GetInstances and GetImages on aws and openstack (#660)

fabioDMFerreira:


lepton/openstack.go
```diff
@@ -354,19 +339,52 @@ func (o *OpenStack) ListInstances(ctx *Context) error {
...
+				Status:    s.Status,
+				Created:   s.Created.Format("2006-01-02 15:04:05"),
+			}
 
+			cinstances = append(cinstances, cinstance)
 		}
-		table.Render()
 
-		return false, nil
+		return true, nil
```
fabioDMFerreira, line 372: Replaced by true to iterate over all pages. The value false makes the iterator end prematurely.

@eyberg confirm whether reading only the first page is intentional.

eyberg: nope - just a bug - thanks!

eyberg (approved):

PR #662, opened 2020-10-04
allow mounting volume by label

ronaudinho:
1. json removed from volumes dir, `ops volume list` for onprem now scans default volume dir instead
2. refactor `lepton/volume.go` to `lepton/onprem_volume.go` and create volume stubs for each cloud provider
3. to label volume locally, creates symlink. mount related operations will resolve the symlink and points to actual file so there is no logic change in mount here
4. skipped attach/detach for local since these operations are achieved on run (by adding args to `qemu` command). use `ops run/load -t onprem --mounts <label/uuid>:<mount> ...` to test running image with mounted volume
5. use `ops image --mounts <label/uuid>:<mount>` to create cloud images with paths for mounting, `ops volume attach/detach` for these image in next PR


ronaudinho:
@eyberg @francescolavra any chance you have `mkfs` binary before this https://github.com/nanovms/nanos/pull/1303?
i tested mounting using [empty](https://github.com/nanovms/ops-examples/commit/560e3b641b60197ebe657c11068a22375f6ff16b#diff-1a1dac75901b942d16b53a3a31140f90) and [non-empty](https://github.com/nanovms/ops-examples/commit/560e3b641b60197ebe657c11068a22375f6ff16b#diff-4dcd70db55958f7b4af8d70bc328dc55) volumes but it does not seem to work now, need to validate if they are related to my changes

i am using `mkfs` nightly since manually building nanos still don't work on my machine.

cmd/volume.go
```diff
@@ -76,20 +99,71 @@ func volumeListCommand() *cobra.Command {
...
+
+	err = vol.UpdateVolume(conf, name, label)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func volumeUpdateCommand() *cobra.Command {
+	cmdVolumeUpdate := &cobra.Command{
+		Use:   "update <volume_name:volume_uuid> <new_label>",
```
francescolavra, line 130: I don't know if there are specific requirements to allow a user to update the label of an existing volume, but this won't work: the label specified when creating a volume with mkfs cannot be changed later on (it's encoded in the .raw file). You can change the symlink for an on-prem volume file, but its internal label will remain the same, so you won't be able to use the new label in a `mounts` directive in a manifest.

lepton/onprem_volume.go
```diff
@@ -0,0 +1,425 @@
...
+		if err != nil {
+			return vol, err
+		}
+		defer src.Close()
+		mkfsCommand.SetStdin(src)
+	}
+
+	mkfsCommand.SetupCommand()
+	err := mkfsCommand.Execute()
+	// fmt.Printf("mkfs %s\n", strings.Join(mkfsCommand.GetArgs(), " "))
```
eyberg, line 81: cruft

cmd/volume.go
```diff
@@ -122,20 +208,21 @@ func volumeAttachCommand() *cobra.Command {
...
 		Use:       "volume",
 		Short:     "manage nanos volumes",
-		ValidArgs: []string{"create, list, delete, attach"},
+		ValidArgs: []string{"create, list, update, delete, attach"},
 		Args:      cobra.OnlyValidArgs,
 	}
 	cmdVolume.PersistentFlags().StringVarP(&config, "config", "c", "", "ops config file")
-	cmdVolume.PersistentFlags().StringVarP(&provider, "provider", "p", "", "cloud provider")
+	cmdVolume.PersistentFlags().StringVarP(&label, "label", "l", api.DefaultVolumeLabel, "volume label")
+	cmdVolume.PersistentFlags().StringVarP(&provider, "provider", "p", "onprem", "cloud provider")
```
eyberg, line 221: i'm not married to the short flags but for now the convention looks like '-t' instead of '-p' in most places

eyberg:
this is failing to create a non-empty volume

eyberg (changes requested):
non-empty volumes aren't being filled w/content

fabioDMFerreira:
@ronaudinho creating a non-empty volume wasn't working due to a commit of last week (b450f2447d780478a653a6244662acef1ad6a743).

Soon I'll send a fix for this.

cmd/volume.go
ronaudinho: oh yeah it makes sense, will remove this

ronaudinho:
@fabioDMFerreira thanks, I missed this
@francescolavra nightly works, just verified by creating volume using `mkfs` manually, thanks for the old binary though
@eyberg I can update this PR after #666 is merged

fabioDMFerreira:
I am not able to mount volumes neither. It returns the next error
`mkfs:parse error unknown property discriminator 47`

cmd/volume.go
ronaudinho: updated the flag

ronaudinho:
rebased to #666 and tested `ops run --mounts` with both empty and non-empty volume, by UUID and label, should work locally now

ronaudinho:
> I am not able to mount volumes neither. It returns the next error
> `mkfs:parse error unknown property discriminator 47`

my guess the bug might be in the build manifest process which sanitizes `/` in specified mount path i.e `--mounts id:mnt` and `--mounts id:/mnt` should allow mounting in `/mnt` dir inside nanos

fabioDMFerreira:
@ronaudinho thanks. That was exactly the issue. I added an unnecessary forward slash on the end of mount path `--mounts id:/mnt/`.

I pushed a new commit to this PR. It solves deleting issues: 
* Remove volume file and symlink;
* Delete by UUID and label.

And uses volume name as the label to simplify its usage.

ronaudinho (commented):
@fabioDMFerreira if i understands this correctly, `ops create volume empty` will create `empty:uuid.raw` and symlinks it to `empty.raw`?

PR #669, opened 2020-10-06
Fixed happy path flow now SG gets created in default VPC. Also fixed …

chaitanya-baraskar:
…389.

### DIFF
--- a/cmd/instance.go
+++ b/cmd/instance.go
@@ -39,7 +39,7 @@ func instanceCreateCommandHandler(cmd *cobra.Command, args []string) {
 	}
 
 	zone, _ := cmd.Flags().GetString("zone")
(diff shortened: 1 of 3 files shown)


chaitanya-baraskar:
In PR I fixed following problems.
1. CLI wasn't picking up zone as given condition was checking for null value.
2. Added fix for 389 issue now it just shows invalid zone doesn't print complete stack trace. To fix this issue I created separate util.go file for lepton package and added exitWithError function. 
3. Previously code was picking up first VPC from multiple VPCs to create SG which is wrong as while creating instance AWS SDK uses default VPC if you don't mention one. So I added code to retrieve default VPC from list. Current code works fine with new account as it contains only default VPC. In future, we should accept VPC from user if not given we can use default one I have added TODo for same in code.

PR #670, opened 2020-10-06
Fixing 389 and other issues

chaitanya-baraskar:
In PR I fixed following problems.

1. CLI wasn't picking up zone as given condition was checking for null value.
2. Added fix for 389 issue now it just shows invalid zone doesn't print complete stack trace. To fix this issue I created separate util.go file for lepton package and added exitWithError function.
3. Previously code was picking up first VPC from multiple VPCs to create SG which is wrong as while creating instance AWS SDK uses default VPC if you don't mention one. So I added code to retrieve default VPC from list. Current code works fine with new account as it contains only default VPC. In future, we should accept VPC from user if not given we can use default one I have added TODo for same in code.

### DIFF
--- a/lepton/aws.go
+++ b/lepton/aws.go
@@ -375,7 +375,7 @@ func (p *AWS) SyncImage(config *Config, target Provider, image string) error {
 func (p *AWS) CreateInstance(ctx *Context) error {
 	result, err := getAWSImages(ctx.config.CloudConfig.Zone)
 	if err != nil {
-		return err
+		exitWithError("Invalid zone")
 	}
 
 	imgName := ctx.config.CloudConfig.ImageName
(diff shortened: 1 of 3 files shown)


lepton/aws.go
```diff
@@ -472,7 +472,15 @@ func (p *AWS) CreateSG(ctx *Context, svc *ec2.EC2, imgName string) (string, erro
 	if len(result.Vpcs) == 0 {
 		fmt.Println("No VPCs found to associate security group with.")
 	}
-	vpcID := aws.StringValue(result.Vpcs[0].VpcId)
+
+	var vpcID string
+	//TODO: This will fail if there is no default VPC. Need to implement feature where user should be able to mention VPC
+	for i, s := range result.Vpcs {
+		isDefault := *s.IsDefault
```
eyberg, line 479: maybe just default to the first one if no default? although iirc that change went into affect something like 7 years ago; can add an issue later to specify vpc to place in

lepton/utils.go
```diff
@@ -0,0 +1,11 @@
+package lepton
+
+import (
+	"fmt"
+	"os"
+)
+
+func exitWithError(errs string) {
```
eyberg, line 8: this is a dupe but it's relatively innocuous - i suppose there is room for a utils pkg or something in the future

eyberg (approved):

lepton/aws.go
chaitanya-baraskar: The "runinstanceinput" needs VPC if we don't specify one it goes back to default. 

* [EC2-VPC] If you don't specify a subnet ID, we choose a default subnet
from your default VPC for you. If you don't have a default VPC, you must
specify a subnet ID in the request.* [link](https://docs.aws.amazon.com/sdk-for-go/v2/api/service/ec2/#Client.RunInstancesRequest). Chossing first one will result in same problem we are trying to fix.

lepton/utils.go
chaitanya-baraskar: There's a utils in cmd package but I thought lepton package shouldn't depend on cmd. As you said definately in future if we need more utilities like this, I will create separate utils package and refactor code accordingly.

Issue #675, opened 2020-10-07
Default target for all ops commands should be "onprem"


chaitanya-baraskar:
Most of the options have gcp as default. It should be changed to onprem as we don't know user's setup.

PR #679, opened 2020-10-08
Issue - #675

chaitanya-baraskar:
Making "onprem" as default value to -t flag


eyberg:
thanks @chaitanya-baraskar - if you use "fixes #675" github will auto-close tkts that get merged in

PR #699, opened 2020-10-12
Unique openstack instance name.

chaitanya-baraskar:
Openstack Instance name is unique for further instance activities. Added ID column instance list for.


chaitanya-baraskar:
If instance IP address was not available code used to fail with panic. In this PR fixed that as well.

eyberg (approved):

PR #700, opened 2020-10-12
Separate printing logs and getting logs concerns (closes #657)

### LINKED ISSUE #657: GetInstanceLogs in Provider interface should return logs as string or other format
The GetInstanceLogs in every cloud provider struct is printing the logs in the console instead of returning the logs to the caller.

The printing should be handled by the method caller.

fabioDMFerreira:
* Add safeguards to azure getIstances

* Add safeguard to openstack getInstances

### DIFF
--- a/lepton/aws.go
+++ b/lepton/aws.go
@@ -672,8 +672,18 @@ func (p *AWS) DeleteInstance(ctx *Context, instancename string) error {
 	return nil
 }
 
+// PrintInstanceLogs writes instance logs to console
+func (p *AWS) PrintInstanceLogs(ctx *Context, instancename string, watch bool) error {
+	l, err := p.GetInstanceLogs(ctx, instancename)
+	if err != nil {
+		return err
+	}
+	fmt.Printf(l)
+	return nil
+}
+
 // GetInstanceLogs gets instance related logs
-func (p *AWS) GetInstanceLogs(ctx *Context, instancename string, watch bool) error {
+func (p *AWS) GetInstanceLogs(ctx *Context, instancename string) (string, error) {
 	svc, err := session.NewSession(&aws.Config{
 		Region: aws.String(ctx.config.CloudConfig.Zone)},
 	)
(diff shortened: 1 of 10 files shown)


lepton/aws.go
```diff
@@ -672,8 +672,18 @@ func (p *AWS) DeleteInstance(ctx *Context, instancename string) error {
 	return nil
 }
 
+// PrintInstanceLogs writes instance logs to console
+func (p *AWS) PrintInstanceLogs(ctx *Context, instancename string, watch bool) error {
+	l, err := p.GetInstanceLogs(ctx, instancename)
```
sanketsudake, line 677: @fabioDMFerreira I just wonder why do we need to it this way. I see across all providers we have the same implementation for the method. Should we just handle it at `cmd/instance.go`?
Ideally, we should avoid printing logs from providers or common logs. Consumers of the method can decide if to print logs or not.

lepton/openstack.go
```diff
@@ -390,10 +392,16 @@ func (o *OpenStack) GetInstances(ctx *Context) ([]CloudInstance, error) {
 		for _, s := range serverList {
 			// fugly
 			ipv4 := ""
+<<<<<<< HEAD
```
sanketsudake, line 395: Some conflicts here, can you please rebase your PR?

fabioDMFerreira: thanks for the heads up @sanketsudake

lepton/aws.go
fabioDMFerreira: I did the same question while coding @sanketsudake.

We have a `watch` flag (still not supported in any provider) in the command that should listen for new changes in the log and print them in real-time.  This implementation may be different from provider to provider.

Ideally, we should return a read stream, but for simplicity reasons, we're returning a string.

fabioDMFerreira: Created a new ticket for this #702.

eyberg: I think the main reason for keeping it in provider specific methods is cause of the tail/watch functionality, gcp has this functionality && vsphere could have it, I think azure has support for it too, so prob. worth keeping as-is even if it seems a bit redundant for now

eyberg (approved):

Issue #625, opened 2020-09-23
openstack image names are not unique


eyberg:
* meaning we can reuse the name for timestamping
* delete image operations need to reflect this

chaitanya-baraskar:
I have started working on this one. I will fix unique name issue while fixing delete instance flow.

PR #703, opened 2020-10-13
Issue #625

chaitanya-baraskar:
Removing default value for flavor flag
Added support for deleting openstack instance


cmd/instance.go
```diff
@@ -90,7 +90,7 @@ func instanceCreateCommand() *cobra.Command {
 
 	cmdInstanceCreate.PersistentFlags().StringVarP(&config, "config", "c", "", "config for nanos")
 	cmdInstanceCreate.PersistentFlags().StringVarP(&imageName, "imagename", "i", "", "image name [required]")
-	cmdInstanceCreate.PersistentFlags().StringVarP(&flavor, "flavor", "f", "g1-small", "flavor name for GCP")
```
eyberg, line 93: this should be the default for gcp - i'm ok w/it not being the default here but we should set it as default if empty string for gcloud

chaitanya-baraskar: In case user doesn't provide any value, for openstack ops selects one flavour and deploys instance. In case we want to provide default value we should hardcode it in cloud provider code e.g. [gcp.go ](https://github.com/nanovms/ops/blob/master/lepton/gcp.go)As default for every cloud provider could be different and in case of openstack we can continue following same approach as flavor name varies per installation. @eyberg please let me know which approach we should follow.

eyberg: yes, every provider should have a default and it's ok to hard-code those in their respective locations for now

Issue #701, opened 2020-10-12
ops should warn user if running inside a vm with accel enabled


eyberg:
i thought this was fixed but someone just reported running into this

```
You specified hardware acceleration, but it is not supported
```

should probably ask the pointed question "Are you running inside a vm? If so disable accel with --accel=false"

PR #707, opened 2020-10-13
change warning message (#701)

emikhalev:


eyberg (commented):
thanks - I'd prob. just keep both of these lines though separated by newline - the second one is more of a hint but not necessarily accurate

eyberg:
awesome - thanks!

Issue #705, opened 2020-10-13
Expose list of ports on instances


chaitanya-baraskar:
Right now only port 80 and 443 are exposed on instance. Ports should be exposed as per config.

eyberg:
pretty sure this is https://github.com/nanovms/ops/issues/683

chaitanya-baraskar:
yes... closing issue... #709

PR #709, opened 2020-10-14
Expose list of ports #705

chaitanya-baraskar:
Issue #705 

-Added support to expose list of ports given by user.
-Removed http and https ports as default ports.
-In case user doesn't provide flavour, added default flavour "g1-small"


eyberg (approved):
this is fine once govet is fixed

PR #711, opened 2020-10-16
Issue #625 - Removing default value for flavor flag

chaitanya-baraskar:
Issue #625 

Removing default value for flavor flag
Added support for deleting openstack instance
Added default flavor for aws

### DIFF
--- a/cmd/instance.go
+++ b/cmd/instance.go
@@ -90,7 +90,7 @@ func instanceCreateCommand() *cobra.Command {
 
 	cmdInstanceCreate.PersistentFlags().StringVarP(&config, "config", "c", "", "config for nanos")
 	cmdInstanceCreate.PersistentFlags().StringVarP(&imageName, "imagename", "i", "", "image name [required]")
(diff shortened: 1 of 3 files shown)


cmd/instance.go
```diff
@@ -90,7 +90,7 @@ func instanceCreateCommand() *cobra.Command {
 
 	cmdInstanceCreate.PersistentFlags().StringVarP(&config, "config", "c", "", "config for nanos")
 	cmdInstanceCreate.PersistentFlags().StringVarP(&imageName, "imagename", "i", "", "image name [required]")
-	cmdInstanceCreate.PersistentFlags().StringVarP(&flavor, "flavor", "f", "g1-small", "flavor name for GCP")
```
eyberg, line 93: again, can you set g1-small as the default in gcloud if not specified here please?

eyberg (changes requested):

eyberg (approved):

cmd/instance.go
eyberg: this was resolved in a diff. pr https://github.com/nanovms/ops/pull/709/files#diff-7a003c5767841ecce9f30f3795aee8eb0b854a61fcf1b4efd04357d262ec883dR353

Issue #672, opened 2020-10-06
ops image list fails on invalid region


eyberg:
same problem as https://github.com/nanovms/ops/issues/389

chaitanya-baraskar:
once you accept #389 this will be quick fix as I will be able to use utils.

chaitanya-baraskar:
There's no panic in code any more and CLI is giving sufficient infromation right now. This issue is resolved.

eyberg:
I believe you're referring to 389 ?

this one looks like it still needs an error msg along the lines of "Invalid Region"

```
➜  ~ ops image list -t aws -z asdf
RequestError: send request failed
caused by: Post "https://ec2.asdf.amazonaws.com/": dial tcp: lookup ec2.asdf.amazonaws.com: no such host
➜  ~ ops image list -t aws -z us-west-2
+------+----+--------+---------+
| NAME | ID | STATUS | CREATED |
+------+----+--------+---------+
```

PR #714, opened 2020-10-19
Bugfix 672 - invalid region message

chaitanya-baraskar:
Issue - #672

Fixed invalid region message for aws in "ops instance list"


eyberg:
this fixes it for ops instance list but not ops image list which is #672 .... anyways - merging this one

Issue #718, opened 2020-10-21
Auto-load DNS record on launching an instance


fabioDMFerreira:
We should be able to provide a flag that automatically registers the domain name record with the launched instance IP in the cloud provider.

eyberg:
I think this works now w/the domain flag - at least for gcp

PR #729, opened 2020-10-23
Add openstack DNS operations

fabioDMFerreira:
#718

### DIFF
--- a/lepton/openstack_dns.go
+++ b/lepton/openstack_dns.go
@@ -1,0 +1,106 @@
+package lepton
+
+import (
+	"os"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/dns/v2/recordsets"
+	"github.com/gophercloud/gophercloud/openstack/dns/v2/zones"
+)
+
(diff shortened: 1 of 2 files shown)


lepton/openstack_dns.go
```diff
@@ -0,0 +1,106 @@
...
+
+	allZones, err := zones.ExtractZones(allPages)
+	if err != nil {
+		return "", err
+	}
+
+	if len(allZones) == 0 {
+		createOpts := zones.CreateOpts{
+			Name:  dnsName + ".",
+			Email: "martinhoferreira10@gmail.com",
```
eyberg, line 37: ? is this required or ?

eyberg (approved):
just need to fix the email

lepton/openstack_dns.go
fabioDMFerreira: I added my personal email by accident. Without email parameter, it isn't possible to create a zone in openstack.

eyberg: that seems weird? if you need something you can just use the example.com from http://www.faqs.org/rfcs/rfc2606.html

fabioDMFerreira: I added a default email with format `admin@<domain name`. I guess it makes sense the user to add a command flag that sets his email.

Issue #715, opened 2020-10-20
add interactive debugger support


eyberg:
right now debugging is not super intuitive - we should do the following:

- [x] have a switch to pop an interactive debugger on locally (-s -S)
- [x] pop gdb listener open on running qemu
- [x] have flag to rebuild image w/aslr off  (```m.AddDebugFlag("noaslr", 't')```)
- [x] prompt user if binary is not statically linked
- [x] prompt user if symbols aren't available (stripped or just not even compiled in)
- [x] force accel off (--accel=false)
- [x] give instructions to user to add symbols && connect remotely
- [x] consider moving '-d to something like '-trace' (although -t is taken for region/zone); consider this flag being '-debug'; maybe '-t' should be '-r or -z' ?? not even sure why that is '-t'

~~(also give instructions for end-user to connect w/gdb)~~
*we might as well pop gdb open and issue the connect/symbol load as well to prevent any confusion*

(for the last 2 we'll need to adjust this process if keeping aslr on or if dynamically linked)

eyberg:
tmp instructions https://nanovms.gitbook.io/ops/debugging

PR #733, opened 2020-10-28
Add flags to allow interactive debugging #715

fabioDMFerreira:
* Disable accel;

* Append noaslr flag;

* Check program is statically linked and has debugging symbols.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -160,6 +160,8 @@ func runCommandHandler(cmd *cobra.Command, args []string) {
 	if debug {
 		c.RunConfig.Debug = true
 
+		c.Debugflags = append(c.Debugflags, "noaslr")
+
 		elfFile, err := api.GetElfFileInfo(c.ProgramPath)
 		if err != nil {
 			log.Fatal(err)
@@ -172,10 +174,6 @@ func runCommandHandler(cmd *cobra.Command, args []string) {
 		if !api.HasDebuggingSymbols(elfFile) {
 			log.Fatalf("Program %s must be compiled with debugging symbols", c.ProgramPath)
 		}
-	}
-
-	if trace || debug {
-		c.Debugflags = append(c.Debugflags, "noaslr")
 	}
 
 	c.RunConfig.GdbPort = gdbport
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -41,7 +41,12 @@ func runCommandHandler(cmd *cobra.Command, args []string) {
 		panic(err)
 	}
 
-	debugflags, err := strconv.ParseBool(cmd.Flag("debug").Value.String())
+	debug, err := strconv.ParseBool(cmd.Flag("debug").Value.String())
+	if err != nil {
+		panic(err)
+	}
+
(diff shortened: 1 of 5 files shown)


eyberg:
the code here is fine, however, the syscall tracing needs to be separated as it vomits output, espc in larger applications whereas this functionality is more of an interactive step-by-step/breakpoint/watchpoint/etc..

I guess I'm open to suggestion on what we should do here but I think the syscall tracing should be put into a diff flag something and this turn into the actual 'debug' flag - but they don't belong together for sure

if anyone from @nanovms/engineering has comments feel free to chime in

eyberg:
also - maybe it's worth popping some instructions to the end-user or a link to https://nanovms.gitbook.io/ops/debugging

francescolavra:
I agree that debug and tracing should be two separate options.
I guess we could use --trace (with no shorthand synonym, because -t is already used to specify the target cloud platform) to enable kernel tracing, and -d / --debug to enable interactive debugging.
As a side note, I may be missing something but I don't see why acceleration would need to be disabled in order to allow interactive debugging. Maybe @eyberg can shed some light on why forcing accel off is in the checklist at https://github.com/nanovms/ops/issues/715?

wjhun:
> As a side note, I may be missing something but I don't see why acceleration would need to be disabled in order to allow interactive debugging. Maybe @eyberg can shed some light on why forcing accel off is in the checklist at #715?

AFAIK, debugging via the qemu gdb stub only works when qemu is running with TCG.

francescolavra:
I tested the gdb stub with KVM enabled, and it sort of works, but with (at least) one major inconvenience: it's not possible to insert breakpoints in the kernel until memory paging has been enabled. I believe this is due to the fact that with KVM enabled, GDB cannot use software breakpoints and must use hardware breakpoints instead.
Without acceleration this problem does not appear. So it makes perfect sense for Ops to force accel off when the debug flag is set.

eyberg:
@fabioDMFerreira - so if it's possible to have the syscall trace be '--trace' (and keep -t for cloud selection for now), and then convert '-d' to actual debugger let's do that and then we'll merge this

cmd/run.go
```diff
@@ -141,10 +151,33 @@ func runCommandHandler(cmd *cobra.Command, args []string) {
...
+		if api.IsDynamicLinked(elfFile) {
+			log.Fatalf("Program %s must be linked statically", c.ProgramPath)
+		}
+
+		if !api.HasDebuggingSymbols(elfFile) {
+			log.Fatalf("Program %s must be compiled with debugging symbols", c.ProgramPath)
+		}
+	}
+
+	if trace || debug {
```
eyberg, line 177: this is just for debug not for trace

eyberg:
this is failing to build on osx cause the folowing is only defined in ldd_linux - the underlying lib might only build in linux so i'm ok stubbing this out for now in ldd_darwin but there's no reason why they shouldn't be able to work under osx

eyberg:
excellent - thanks!

Issue #388, opened 2019-08-19
aws ami creation could use a progressbar


eyberg:
the upload is fairly fast but the creation of an ami can take like 5 minutes to do - not sure why -anyways there is a waiter there already - i'm pretty sure it can be refactored to ping the status every 10-20 seconds with a more up-to-date time estimate for a spinner

chaitanya-baraskar:
I will do some research on this topic.

PR #737, opened 2020-10-28
Issue 388

chaitanya-baraskar:
Issue - #388 

Added progress bar for better user experience.
Created separate utils package to remove interdependency and refactored code accordingly.


cmd/image.go
```diff
@@ -2,10 +2,13 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/briandowns/spinner"
```
eyberg, line 5: i'm not opposed to using a new spinner but don't we have an existing one?

utils/utils.go
```diff
@@ -7,20 +7,27 @@ import (
...
+	ErrorColor = "\033[1;31m%s\033[0m"
+)
+
+func ExitWithError(errs string) {
 	fmt.Println(fmt.Sprintf(ErrorColor, errs))
 	os.Exit(1)
 }
 
-func arrayToString(a []int, delim string) string {
+func ArrayToString(a []int, delim string) string {
```
eyberg, line 22: govet will complain here

eyberg (changes requested):
I don't care which one you pick but I think we are already using https://github.com/schollz/progressbar/ as a progressbar - i'd just do one or the other - no need for 2, everything else looks fine

chaitanya-baraskar:
@eyberg this is how spinner works https://asciinema.org/a/zANzbg2cnEieFGyVxnmAg2vAY

eyberg:
ok, well in that case maybe we should just use the existing progressbar - the whole point of that tkt was to provide an indicator of progress (25% done, 50% done, etc.) - there are facilities in the aws sdk to show progress for things like that

chaitanya-baraskar:
Ok. I am not aware of this feature from SDK let me do some research.

PR #744, opened 2020-11-04
Create firewall rules with UDP protocol if specified in Azure (closes #742)

### LINKED ISSUE #742: azure NSG creation needs to unstub TCP and allow UDP if specified
https://github.com/nanovms/ops/blob/master/lepton/azure_network.go#L293

fabioDMFerreira:


lepton/azure_network.go
```diff
@@ -285,12 +285,19 @@ func (a *Azure) CreateNetworkSecurityGroup(ctx context.Context, location string,
 
 	securityRules := make([]network.SecurityRule, len(c.RunConfig.Ports))
 
+	var communicationProtocol network.SecurityRuleProtocol
+	if c.RunConfig.UDP {
```
eyberg, line 289: i think the convention we are using right now is tcp && udp not tcp || udp - maybe this should change in the future but prob. best to keep the same for now

fabioDMFerreira: prob. we should use flags to distinguish the ports which use tcp or udp (--tcp and --udp) instead of a configuration variable

eyberg: i'm in favor of adding a --udp flag, although tcp is 99% going to be the default so I think for now we should just keep it as an addition, so:

1) make this be tcp+udp if specified
2) add support for --udp for ease of use

fabioDMFerreira: We're using `-p` in some commands to provide the configuration value for cloud provider project. I think `ops instance create` is the only one which uses `-p` to defined the instance port. I know it may be difficult to keep the consistency across all commands, but we can use this opportunity to change the port flag.

fabioDMFerreira:
@eyberg also changed aws code to support tcp or udp ports simultaneously.

fabioDMFerreira:
if we change https://github.com/nanovms/ops/blob/master/lepton/qemu.go#L260 to use the new run configuration property `UDPPorts` we can deprecate `UDP` in the running configuration.

eyberg (approved):

PR #758, opened 2020-11-12
Allow ops usage inside an unikernel

fabioDMFerreira:
Use current directory if there is no way of getting user home directory.

### DIFF
--- a/lepton/homedir.go
+++ b/lepton/homedir.go
@@ -3,6 +3,7 @@ package lepton
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


eyberg (approved):
that dirUnix function looks overly complicated for no reason but this seems fine to me

fabioDMFerreira:
I simplified and removed unused methods from homedir file.

The removed code is based on https://github.com/mitchellh/go-homedir. It has alternative methods to get the home directory in case of HOME environment variable not being present.

The current method uses the go native method https://golang.org/pkg/os/#UserHomeDir.

Issue #400, opened 2019-10-31
optional debug loglevel


eyberg:
there's lots of useful debug output that gets thrown away w/underscores - we might want to craft in a debugging output loglevel rather than trash the output of some of the requests

eyberg:
there's also a lot of random debug output that is currently being cat out that probably doesn't belong unless a debug loglevel is shown

eyberg:
in addition to debug log level it would prob. be good to have a warning log level as well

we have quite a few warnings (which some ppl might wish to have silenced) but prob. should be on by default and then we have quite a few places in the code where it would be worthwhile to have debug output that should only be turned on at the bequest of a dev

fabioDMFerreira:
There are two libraries we can use:
* https://github.com/sirupsen/logrus
* https://github.com/uber-go/zap

We can also create our own logging module.

eyberg:
I wasn't really looking for anything extensive - just a way to separate warnings from errors from debug output

PR #762, opened 2020-11-13
Allow filtering log messages per level (#400)

fabioDMFerreira:
* Add logger;

* Add command flags "show-error" and "show-warnings" to every command;

* Append logger to context;


fabioDMFerreira:
In this PR, we append a logger utility to the context. This way we can use it in every ops command.

The exceptions are the volumes operations which don't have access to a context (https://github.com/nanovms/ops/issues/763).

With this, we can filter the messages printed in the user console (https://github.com/nanovms/ops/issues/764).

cmd/appendGlobalCmdFlagsToConfig.go
```diff
@@ -0,0 +1,12 @@
+package cmd
+
+import (
+	"github.com/nanovms/ops/lepton"
+	"github.com/spf13/pflag"
```
eyberg, line 5: is it necessary to have another flag lib? we already have one in addition to the existing stdlib stuff

lepton/logger_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/nanovms/ops/lepton"
+)
+
+const (
+	colorRed    = "\033[31m"
```
eyberg, line 11: why are these repeated from lepton/logger?

eyberg:
the only concern here is the use of yet another flag lib - it's not possible to consolidate?

eyberg (approved):

cmd/appendGlobalCmdFlagsToConfig.go
fabioDMFerreira: This is an internal library used by `spf13/cobra`. We're using to receive `cmd.Flags()` result in a function.

lepton/logger_test.go
fabioDMFerreira: didn't make sense to me making these constants values readable from external packages.

PR #772, opened 2020-11-17
750 delete ips after deleting azure instances

fabioDMFerreira:


lepton/azure.go
```diff
@@ -386,74 +409,70 @@ func (a *Azure) CreateInstance(ctx *Context) error {
...
 	// create virtual network
+	ctx.logger.Info("creating virtual network with id %s\n", vmName)
 	vnet, err := a.CreateVirtualNetwork(context.TODO(), location, vmName)
 	if err != nil {
-		fmt.Println(err)
-	}
-
-	if debug {
-		fmt.Printf("%+v\n", vnet)
+		ctx.logger.Error("error creating virtual network: %v\n", err)
```
fabioDMFerreira, line 427: The idea here is to display error details if debug flag is enabled and return the short version of the error to the caller.

We're introducing a lot of repetition this way. We could create an entity responsible for logging errors and create errors.

lepton/azure.go
```diff
@@ -106,20 +107,26 @@ func (a *Azure) GetResourceManagementAuthorizer() (autorest.Authorizer, error) {
 	return armAuthorizer, err
 }
 
-func (a *Azure) getImagesClient() compute.ImagesClient {
+func (a *Azure) getImagesClient() (*compute.ImagesClient, error) {
 	vmClient := compute.NewImagesClientWithBaseURI(compute.DefaultBaseURI, a.subID)
-	authr, _ := a.GetResourceManagementAuthorizer()
+	authr, err := a.GetResourceManagementAuthorizer()
```
fabioDMFerreira, line 112: the resource management authorizer is required in every azure client (images, vms, virtual networks, etc). 

The handling of errors every time we get a new client adds boilerplate and in some situations unnecessary code repetitions.

lepton/azure_network.go
```diff
@@ -69,20 +89,157 @@ func (a *Azure) CreateNIC(ctx context.Context, location string, vnetName, subnet
...
+	nsgClient, err := a.getNsgClient()
+	if err != nil {
+		return err
+	}
+	subnetsClient, err := a.getSubnetsClient()
+	if err != nil {
+		return err
+	}
+	vnetClient, err := a.getVnetClient()
+	if err != nil {
```
fabioDMFerreira, line 166: This is an example of boilerplate (error handling) and unnecessary code executed referred in a previous comment.

`getNsgClient`, `getSubnetsClient` and `getVnetClient` requires a resource management authorizer. Each one of these methods is doing a network call to get it.

cmd/appendGlobalCmdFlagsToConfig.go
```diff
@@ -9,4 +9,5 @@ import (
 func AppendGlobalCmdFlagsToConfig(cmdFlags *pflag.FlagSet, config *lepton.Config) {
 	config.RunConfig.ShowWarnings, _ = cmdFlags.GetBool("show-warnings")
 	config.RunConfig.ShowErrors, _ = cmdFlags.GetBool("show-errors")
+	config.RunConfig.Debug, _ = cmdFlags.GetBool("debug")
```
eyberg, line 12: doesn't this conflict with existing '-d' / '--debug' flag? - maybe follow your convention of 'show-debug' -- or we could take the increasingly verbosity approach - we have '-v' - could have '-vv' and '-vvv' - or could change all of this to '-log-level=' ??

lepton/azure.go
eyberg: errors need to be handled regardless but if it's used by a common set of methods could potentially abstract it out

lepton/azure.go
eyberg: +1

eyberg (approved):
just need to look at '-debug' conflicts

cmd/appendGlobalCmdFlagsToConfig.go
fabioDMFerreira: There are no conflicts. The debug flag allows the interactive debugger and displays the debug messages. The only minor issue is that it overrides the flag description message.

But sounds good trying other flags approach. I was thinking of deprecating the `--show-errors` and `--show-warnings`. I like the `-v`,`-vv` and `-vvv` approach.
-v -> show errors
-vv -> show errors and information messages
-vvv -> show errors, warnings and information messages

lepton/azure.go
fabioDMFerreira: https://github.com/nanovms/ops/issues/778

lepton/azure_network.go
fabioDMFerreira: https://github.com/nanovms/ops/issues/778

lepton/azure.go
fabioDMFerreira: https://github.com/nanovms/ops/issues/779

cmd/appendGlobalCmdFlagsToConfig.go
eyberg: i'm ok w/that - just to be clear the debug flag should be a halt-the-world && launch the gdb listener - this is more for debugging api integration via debug log statements and these need to be 2 separate concerns, which is why I'm in favor of having a different naming convention of some kind, although this is kind of out of scope of this tkt so as long as there is no conflict (between the aforementioned) feel free to merge

fabioDMFerreira: I'm adding the new flag `show-debug` to keep consistency with the names of the existing flags. We can review this on another ticket.

PR #785, opened 2020-11-20
Allow specify vpc, subnet and security group on launching azure instances

### LINKED ISSUE #748: user should be able to specify {vnet, nsg, subnet} for instances in azure
along the same lines as this tkt -> https://github.com/nanovms/ops/issues/734

essentially right now we auto-create (which should remain default) all of this stuff per instance and makes things a bit more difficult if you wanted to spin up an instance in the same network as something else

### LINKED ISSUE #784: Azure - allow to reuse a VPC, a Security group and a subnet on launching an instance

fabioDMFerreira:
(closes #748 closes #784)

### DIFF
--- a/lepton/azure.go
+++ b/lepton/azure.go
@@ -421,44 +421,74 @@ func (a *Azure) CreateInstance(ctx *Context) error {
 	ctx.logger.Log("spinning up:\t%s\n", vmName)
 
 	// create virtual network
-	ctx.logger.Info("creating virtual network with id %s\n", vmName)
-	vnet, err := a.CreateVirtualNetwork(context.TODO(), location, vmName)
-	if err != nil {
-		ctx.logger.Error("error creating virtual network: %v\n", err)
-		return errors.New("error creating virtual network")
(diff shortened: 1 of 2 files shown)


lepton/azure.go
```diff
@@ -421,44 +421,74 @@ func (a *Azure) CreateInstance(ctx *Context) error {
...
 	}
 
 	// create nsg
-	ctx.logger.Info("creating network security group with id %s\n", vmName)
-	nsg, err := a.CreateNetworkSecurityGroup(context.TODO(), location, vmName, c)
-	if err != nil {
-		ctx.logger.Error("error creating network security group: %v", err)
-		return errors.New("error creating network security group")
+	var nsg *network.SecurityGroup
+	configSecurityGroup := ctx.config.RunConfig.VPC
```
eyberg, line 444: isn't this supposed to be security group?

eyberg (changes requested):
just s/vpc/sg on L444 I think

Issue #730, opened 2020-10-26
kernel module support


eyberg:
we're implementing a form of kernel modules in nanos - we'll want to expose this functionality in ops once we have the first bit of implementation

eyberg:
https://github.com/nanovms/nanos/pull/1321

eyberg:
this is still a WIP as the telemetry stuff needs to be extracted out into the tls klib https://github.com/nanovms/nanos/pull/1325

eyberg:
https://github.com/nanovms/ops/issues/761

eyberg:
for now what I would probably do is allow ops to consume any klibs found in ~/.ops/klibs && if specified in config

something like ->

```Klibs: ["tls", "radar"]```

still need to figure out if https://github.com/nanovms/ops/issues/761 makes sense - I think for at least public klibs it does make sense

PR #786, opened 2020-11-23
Add kernels modules support (#730)

fabioDMFerreira:


eyberg (changes requested):
instead of a klib bool this should explicitly be told which klibs in the config.json are wanted - there's an expectation that there are going to be quite a few and all are optional

fabioDMFerreira:
Added `Klibs` key to `config.RunConfig`. We're checking whether the specified klibs are in ops klibs directory. If not, we just ignore the klib.

eyberg (approved):

Issue #413, opened 2019-11-07
libnss being overwritten even when local version is specified in config.json ?


eyberg:
@wjhun reported that libnss was being dropped in from the common tarball even when he was overwriting it with 'Files' in config.json which was messing up symbol resolution

I believe this was added at some point in the past to support earlier go versions that wanted it as we by default include this into the various packages we distribute/support, however I just tried w/go 1.12 which is a version behind and it didn't look for it at all

PR #796, opened 2020-11-30
Add default files to manifest before using the configuration ones 

fabioDMFerreira:
#413

Prevent default files from overriding configuration files.


fabioDMFerreira:
The user will see the next warning message if he overrides the default files.

`warning: overwriting existing file lib/x86_64-linux-gnu/libnss_dns.so.2 hostpath old: /home/xyz/.ops/common/libnss_dns.so.2 new: lib/x86_64-linux-gnu/libnss_dns.so.2`

eyberg (approved):

PR #800, opened 2020-12-02
Wait longer for AWS image import

erikmack:
The AWS VM import times out after 15 minutes, but for some reason it requires >20 minutes for me.  Please increase the timeout.


eyberg:
i'm fine upping this limit but 20min is super long - does it actually get created after then?

eyberg (approved):

PR #811, opened 2020-12-03
Allow set instance name (closes #807)

### LINKED ISSUE #807: Allow users to specify instance name
In aws, we're allowing users to override the instance name by using a tag with the key name.

We can add a flag to `ops instance create` command. If not specified we can continue to concatenate the image name with the timestamp.

This change affects every supported cloud provider.

fabioDMFerreira:
* Use command flag or configuration file to set instance name

* Use image name and timestamp to create instance name if not provided

* Update aws, azure, gcp and openstack instance name on creating them

### DIFF
--- a/cmd/instance.go
+++ b/cmd/instance.go
@@ -3,8 +3,10 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	api "github.com/nanovms/ops/lepton"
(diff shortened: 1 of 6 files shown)


fabioDMFerreira:
The instance name can be set using command argument or configuration.

Command example
`go run ops.go instance create walk-server -i walk-server-image`

Configuration example
`{
  "RunConfig":{
    "InstanceName":"walk-server"
  }
}`

eyberg (approved):

PR #849, opened 2020-12-30
(Suggestion): Config structure/order and variable descriptions

FredrikBakken:
In this pull-request, I am providing some updated changes to the configuration file. These suggestive changes are:

- Re-ordering the config variables in alphabetical order, which I am also working on for the "Configuration" page in the `ops-documentation`. This helps any developer/user who wants to use the OPS application by finding the necessary parameters.
- Adding comments to each variable to define what is used for, whether it is optional or not, and what the default value for the variable is. The comment structure is inspired by the config file for the Sarama library: https://github.com/Shopify/sarama/blob/master/config.go

@eyberg, any comments, feedback, or further suggestions for these changes?

### DIFF
--- a/lepton/config.go
+++ b/lepton/config.go
@@ -15,7 +15,7 @@ type Config struct {
 	// BuildDir
 	BuildDir string
 
-	// CloudConfig
+	// CloudConfig configures various attributes about the cloud provider.
 	CloudConfig ProviderConfig
 
 	// Debugflags
@@ -84,7 +84,7 @@ type Config struct {
 
 // ProviderConfig give provider details
 type ProviderConfig struct {
-	// BucketName
+	// BucketName specifies the bucket to store the ops built image artifacts.
 	BucketName string `cloud:"bucketname"`
 
 	// Flavor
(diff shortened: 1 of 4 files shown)


eyberg:
i'm ok w/this change, cc'ing @fabioDMFerreira for his input - the only thing I would say is all the comments saying "to be described" are prob not worth including

FredrikBakken:
Thank you for the feedback, @eyberg. Made some minor updates based on your comments where I have left some of the variables undocumented.

The changes in this pull-request are also related to the pull-request found in the `ops-documentation`: https://github.com/nanovms/ops-documentation/pull/148. Documentation of these variables would be beneficial to have both places since some developers look into the source code directly, while others use the docs.

fabioDMFerreira (approved):
Thanks for this @FredrikBakken! Looks good

lepton/config.go
```diff
@@ -2,79 +2,212 @@ package lepton
...
+
+	// Platform defines the cloud provider to use with the ops CLI, currently
+	// supporting aws, azure, and gcp.
+	Platform string `cloud:"platform"`
+
+	// ProjectID is used to define the project ID when the Platform is set
+	// to gcp.
+	ProjectID string `cloud:"projectid"`
+
+	// Zone is used to define the location of the host resource when the
```
fabioDMFerreira, line 104: Prob. makes sense to add regions sources of other cloud providers. At least AWS and Azure

lepton/config.go
```diff
@@ -2,79 +2,212 @@ package lepton
...
+	// Platform is set to gcp. A list of these zones can be found here:
+	// https://cloud.google.com/compute/docs/regions-zones#available
```
fabioDMFerreira, line 106: Prob. worth to add regions URLs of other cloud providers. At least AWS and Azure

FredrikBakken: Thank you, @fabioDMFerreira - Good point! Added links for AWS and Azure now.

FredrikBakken:
Checked the source code against the Go Report Card (https://goreportcard.com/report/github.com/nanovms/ops) and found a few spelling mistakes that have been fixed and added to this pull-request.

eyberg:
thanks!

PR #870, opened 2021-01-18
Allow to delete old image using a time notation (closes #473)

### LINKED ISSUE #473: add lru image cleaner
note for things like aws we key off of tags to keep multiple deploys of the same application - this should take into account that

fabioDMFerreira:
* Allow to delete more than one image specified on command arguments

* Implement onprem get images method

* Separate concerns of presentation and image getting on onprem provider

* Fix name of aws images list

* Change image cloud "created at" field type

### DIFF
--- a/lepton/onprem_image.go
+++ b/lepton/onprem_image.go
@@ -1,7 +1,6 @@
 package lepton
 
 import (
-	"errors"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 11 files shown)


lepton/onprem_image.go
```diff
@@ -80,18 +66,48 @@ func (p *OnPrem) ListImages(ctx *Context) error {
...
+		tablewriter.Colors{tablewriter.Bold, tablewriter.FgCyanColor},
+		tablewriter.Colors{tablewriter.Bold, tablewriter.FgCyanColor},
+		tablewriter.Colors{tablewriter.Bold, tablewriter.FgCyanColor})
+	table.SetRowLine(true)
+	for _, i := range images {
+		var row []string
+		row = append(row, i.Name)
+		row = append(row, i.Path)
+		row = append(row, bytes2Human(i.Size))
+		row = append(row, i.Created.String())
```
eyberg, line 108: this should be bytes2hHuman

eyberg (changes requested):

eyberg (approved):

Issue #878, opened 2021-01-28
consider adding configurable klib options &&|| exposing manifest entries to user


eyberg:
for instance the ntp klib https://github.com/nanovms/nanos/pull/1385 can take the following vars:

```
ntp_address:pool.ntp.org
ntp_port:123
```

eyberg:
maybe for now as a quick trade we could expose a ManifestVariables that operates similar to the way Env works?

PR #879, opened 2021-01-29
Add ntp_address and ntp_port if ntp klib is specified

fabioDMFerreira:


eyberg (approved):

lepton/manifest.go
```diff
@@ -418,6 +418,28 @@ func (m *Manifest) String() string {
...
+				if val, ok := m.environment["ntpAddress"]; ok {
+					ntpAddress = val
+				}
+
+				if val, ok := m.environment["ntpPort"]; ok {
+					ntpPort = val
+				}
+
+				sb.WriteString(fmt.Sprintf("ntp_address:%s\n", ntpAddress))
+				sb.WriteString(fmt.Sprintf("ntp_port:%s\n", ntpPort))
```
francescolavra, line 437: pool.ntp.org and port number 123 are used by default in the kernel if nothing is specified in the manifest, so maybe I would not put the ntp_address and ntp_port parameters in the manifest unless the user explicitly specifies a value for these options.
As a side note, requiring the user to pollute the environment with parameters that are not used as environment variables isn't ideal, it seems to me it would be better to have such parameters separate from the environment; but that requires the implementation of something like ManifestVariables (https://github.com/nanovms/ops/issues/878).

fabioDMFerreira: The issue with ManifestVariables is the user can pollute the manifest file with whichever variable he wants. Not sure of the implications of this. However, we can limit the variables used in ManifestVariables in ops.

@francescolavra @eyberg any thoughts on this?

eyberg: I see problems with both approaches but I think either way is fine for now - we can refactor this later

PR #886, opened 2021-02-04
add capability of adjusting ntp poll limits

### LINKED ISSUE #885: add capability for user to adjust ntp_poll_min && ntp_poll_max env vars for ntp klib
https://github.com/nanovms/nanos/pull/1394

I think eventually we need to do something klib specific instead of relying on env vars but for now we can just use those to get this out the door

fabioDMFerreira:
(closes #885)

### DIFF
--- a/lepton/manifest.go
+++ b/lepton/manifest.go
@@ -6,6 +6,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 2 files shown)


francescolavra (commented):
Looks good to me

eyberg (approved):

Issue #484, opened 2020-04-02
consider creating a go lib out of mkfs


eyberg:
the manifest generation is super hacky and kinda of an un-needed step, also it'd be cool not to have to shell out for it - although right now mkfs is tightly linked to tfs internals

PR #887, opened 2021-02-06
mkfs: add Go native support for generating image and volume files 

francescolavra:
This change removes the dependency on the external mkfs tool to create image and volume files; serialization of a manifest and
generation of manifest files is therefore not needed anymore, and as such is no longer implemented.
The code to generate image and volume files is being moved to its own package, named "fs"; this new package does not have any
dependency on the other Ops packages, thus there are no circular imports.

Closes https://github.com/nanovms/ops/issues/484.


fabioDMFerreira (approved):

francescolavra:
I updated the cmd/load.go, cmd/run.go, cmd/utils.go, cmd/volume.go, lepton/config.go, lepton/image.go and lepton/onprem_volume_test.go files to remove leftover references to mkfs.

PR #897, opened 2021-02-16
create a new vpc in AWS

### LINKED ISSUE #671: determine vpc preference order
options include:

* rely on default
* create a new vpc
* add option to specify vpc

if your aws acct is over 7 years old you might not have a default - right now the code is just grabbing the very first one it finds, there's a pr to change that to grab the first default or fail

fabioDMFerreira:
(closes #671)

* Add utility to return a cidr block not allocated

* Use vpc name and subnet name to search by vpcs

* Create aws vpc and subnet if names are specified in configuration


fabioDMFerreira:
VPC and Subnet values in AWS configuration must use the resource name instead of the ID. Otherwise, it won't be possible to create an instance in the subnet.

eyberg (approved):

PR #909, opened 2021-02-23
fix: set program in image creation

fabioDMFerreira:


fabioDMFerreira:
The image fix doesn't make sense.

PR #913, opened 2021-02-24
Fix image name used to create an image in openstack 

### LINKED ISSUE #912: fix image creation in openstack
Output:
```
creating image: main
open /home/xyz/.ops/images/main: no such file or directory
exit status 1
```

fabioDMFerreira:
(closes #912)

* Fix volumes name display in openstack

* Remove mandatory bucket from openstack cloud providers

* Fix display of IP addresses on listing ovh instances

* Delete image used to create volume in openstack

### DIFF
--- a/lepton/openstack_instance.go
+++ b/lepton/openstack_instance.go
@@ -37,10 +37,16 @@ func getOpenStackInstances(provider *gophercloud.ProviderClient, opts servers.Li
 
 		for _, s := range serverList {
 			if val, ok := s.Metadata["CreatedBy"]; ok && val == "ops" {
-				// fugly
 				ipv4 := ""
-				// For some instances IP is not assigned.
+
+				// addresses may have a different structure in each cloud provider
+				// vexx
 				z := s.Addresses["public"]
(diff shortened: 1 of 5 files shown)


lepton/openstack_instance.go
```diff
@@ -37,10 +37,16 @@ func getOpenStackInstances(provider *gophercloud.ProviderClient, opts servers.Li
 
 		for _, s := range serverList {
 			if val, ok := s.Metadata["CreatedBy"]; ok && val == "ops" {
-				// fugly
 				ipv4 := ""
-				// For some instances IP is not assigned.
+
+				// addresses may have a different structure in each cloud provider
+				// vexx
```
eyberg, line 43: not ideal but i suppose this is fine for now

eyberg (approved):

lepton/openstack_instance.go
fabioDMFerreira: Yes, openstack should uniformize this. My guess is that one way is openstack API v2 an the other is API v3.

PR #921, opened 2021-03-01
Show correct path of image built for hyper-v

### LINKED ISSUE #920: Fix image path displayed after creating an image with hyper-v

fabioDMFerreira:
closes #920

### DIFF
--- a/cmd/cmd_build.go
+++ b/cmd/cmd_build.go
@@ -51,9 +51,11 @@ func buildCommandHandler(cmd *cobra.Command, args []string) {
 		exitWithError(err.Error())
 	}
 
-	if _, err := p.BuildImage(ctx); err != nil {
+	var imagePath string
+
+	if imagePath, err = p.BuildImage(ctx); err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	fmt.Printf("Bootable image file:%s\n", c.RunConfig.Imagename)
+	fmt.Printf("Bootable image file:%s\n", imagePath)
 }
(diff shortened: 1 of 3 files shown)


cmd/cmd_build.go
```diff
@@ -51,9 +51,11 @@ func buildCommandHandler(cmd *cobra.Command, args []string) {
...
 
-	if _, err := p.BuildImage(ctx); err != nil {
+	var imagePath string
+
+	if imagePath, err = p.BuildImage(ctx); err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	fmt.Printf("Bootable image file:%s\n", c.RunConfig.Imagename)
+	fmt.Printf("Bootable image file:%s\n", imagePath)
```
am11, line 60: Would it make sense to show both the raw image name and the provider-specific image, as both gets created as part of the `ops-build` command? In Hyper-V case, for example:
```sh
$ ops build -t hyper-v myApp
Bootable image file:/home/am11/.ops/images/myApp.img
Hyper-V Disk Image eXtended file: /home/am11/.ops/vhdx-images/myApp.vhdx
```
(name `Hyper-V Disk Image eXtended file` was chosen based on the output of: `file ~/.ops/vhdx-images/myApp.vhdx`)

fabioDMFerreira: Not sure about this. Prob. we should delete the `.img` file because it is only needed to create the `.vhdx` file. What do you think @eyberg?

eyberg: if you are deploying direct I'd be in favor of removing un-used artifacts, many platforms/hypervisors require a certain disk type - many of which we build on-demand, considering how fast these can be made and the fact that even a minute change requires a re-build I don't think it makes a ton of sense to keep them around

fabioDMFerreira: Added commit to remove image artifact. Thanks for your suggestion @am11

eyberg (approved):

Issue #944, opened 2021-03-09
No default IP assigned in hyper-v


am11:
When we run with qemu (e.g. `ops run myapp`), the first line shows up the default IP address added by nanos:

![image](https://user-images.githubusercontent.com/3840695/110471382-c02cea80-80e4-11eb-89e8-a552221bb891.png)

deploying to hyper-v, however, does not assign a default IP address:
![image](https://user-images.githubusercontent.com/3840695/110471572-f10d1f80-80e4-11eb-9e4b-07dc1400d14a.png)

there are *two* 0.0.0.0 and a 127.0.0.1 in hyper-v instance when all interfaces are [queried programmatically](https://docs.microsoft.com/en-us/dotnet/api/system.net.networkinformation.networkinterface.getallnetworkinterfaces); compared to qemu instance, which has 10.0.2.15 and 127.0.0.1 out of the box.

ps: btw, going by [this docs section](https://nanovms.gitbook.io/ops/configuration#runconfig.ip_addr), manually adding `"RunConfig": { "IPAddr": "10.0.2.30"}` to config.json has no effect on qemu or hyper-v instances..

francescolavra:
This might be another issue, but I recently ran into a situation where my Nanos instance (which was previously working perfectly fine) started to fail acquiring the IP address. The network interface of the VM was connected to the hyper-v default switch (i.e. the default setting when creating a new VM). Starting and stopping the hyper-v VM manager Windows service did not help, nor did re-creating the instance from scratch. I could only get it to work again after restarting Windows. I noticed the issue after resuming Windows from hibernation, but I can't say whether hibernation was what triggered the problem.

am11:
Unfortunately, restarting didn't remedy situation on my pc.
tried with simple dotnet console app (https://github.com/nanovms/ops-examples/tree/master/dotnet) with this code:
```c#
using System;
using System.Linq;
using System.Net;
using System.Net.NetworkInformation;
using System.Threading.Tasks;

class Program
{
    static Task Main()
    {
        foreach (NetworkInterface netInterface in NetworkInterface.GetAllNetworkInterfaces())
        {
            Console.WriteLine("Name: {0}, Description: {1}, Addresses: [{2}]",
                netInterface.Name,
                netInterface.Description,
                string.Join(", ",
                    netInterface.GetIPProperties().UnicastAddresses.Select(info => info.Address)));
        }

        Console.WriteLine("Shutdown in 15 sec..");
[...]

PR #946, opened 2021-03-09
Connect to the existing hyper-v virtual switch

am11:
Contributes to #944.


am11:
From `New-VM` cmdlet docs - https://docs.microsoft.com/en-us/powershell/module/hyper-v/new-vm?view=win10-ps

> Hyper-V automatically creates a virtual machine with one virtual network adapter...

so we can just connect to that one in newly created VM rather than adding a new adapter, which confuses nanos.

fabioDMFerreira (approved):

PR #947, opened 2021-03-09
Prepend program name to args after merge

### LINKED ISSUE #942: Issues with args propagation
1. from readme example, if we add `"Args":["one","two","three"]` to config.json, the app receives `two` and `three` but not `one`
2. if we pass args via command line: `ops run -a one -a two -a three ./myapp`, the app does not receive any arguments.
    * only with `ops pkg load ... -a one -a two -a three`, the app receive all three args. but that is not a viable solution for ELF apps..

am11:
Fixes #942

### DIFF
--- a/cmd/flags_build_image.go
+++ b/cmd/flags_build_image.go
@@ -79,6 +79,10 @@ func (flags *BuildImageCommandFlags) MergeToConfig(c *types.Config) (err error) 
 		c.Args = flags.CmdArgs
 	}
 
+	if c.Program != "" {
+		c.Args = append([]string{c.Program}, c.Args...)
+	}
+
 	return
 }
 
--- a/cmd/flags_build_image_test.go
+++ b/cmd/flags_build_image_test.go
@@ -51,6 +51,7 @@ func TestBuildImageFlagsMergeToConfig(t *testing.T) {
 
 	c := &types.Config{
 		VolumesDir: lepton.LocalVolumeDir,
+		Program:    "MyTestApp",
 	}
 	expected := &types.Config{
 		Boot:   opsPath + "/boot.img",
@@ -64,8 +65,8 @@ func TestBuildImageFlagsMergeToConfig(t *testing.T) {
 		RunConfig: types.RunConfig{
 			Imagename: imagesPath + "/test-image.img",
 		},
-		Args:       []string{"a b c d"},
-		Program:    "",
+		Args:       []string{"MyTestApp", "a b c d"},
+		Program:    "MyTestApp",
 		TargetRoot: "unix",
 		Env:        map[string]string{"test": "1234"},
 		NameServer: "8.8.8.8",


fabioDMFerreira (changes requested):
Thanks for this @am11.

We need to handle the args flag in every command that needs to build an image. To avoid setting the program name as the first argument in each command handler we can do it on the struct responsible for merging the build image configuration (https://github.com/nanovms/ops/blob/master/cmd/flags_build_image.go#L65).

If we're using a package configuration to build an image, ops have to merge packages arguments (https://github.com/nanovms/ops/blob/master/cmd/flags_pkg.go#L51). Wondering whether this change will affect the behaviour of `ops image create --package <package> ...` or `ops load <package> ...`

am11:
@fabioDMFerreira, I've addressed the feedback, this is more of a correctness fix. Any suggestions where to put unit test for this so it don't regress?

fabioDMFerreira:
You could add a `Program` here and check whether the final arguments are the expected ones
https://github.com/nanovms/ops/blob/master/cmd/flags_build_image_test.go#L53.

I also think it worth to add some dedicated test to check arguments here
https://github.com/nanovms/ops/blob/master/cmd/flags_pkg_test.go. See next an example of a package configuration.
```
{
  "Program": "mysql_5.7.29/mysqld",
  "Version": "5.7.29",
  "RunConfig": {
    "Memory": "2G"
  },
  "BaseVolumeSz": "200m",
  "Args":["mysqld", "--user=root", "--explicit_defaults_for_timestamp"]
}
```

cmd/flags_build_image.go
```diff
@@ -79,6 +79,8 @@ func (flags *BuildImageCommandFlags) MergeToConfig(c *types.Config) (err error)
 		c.Args = flags.CmdArgs
 	}
 
+	c.Args = append([]string{c.Program}, c.Args...)
```
fabioDMFerreira, line 82: Worth to check if c.Program isn't empty as the program may come from a package.

am11:
It looks like ops-pkg command with current release (0.1.19) does not respect `Args` with `-c config.json`:

```sh
$ cat > test1.js <<<'console.log("args: ", process.argv)'
$ cat > config.json <<<'{ "Args":["one","two","three"] }'

$ ops pkg load node_v11.5.0 -p 8083 -f -n -a test1.js -c config.json
booting /home/am11/.ops/images/test1.js.img ...
You specified hardware acceleration, but it is not supported
Are you running inside a vm? If so disable accel with --accel=false

en1: assigned 10.0.2.15
args:  [ '/node_v11.5.0/node', '/test1.js' ]
exit status 1
```

on the other hand, the command-line args work:

```sh
$ ops pkg load node_v11.5.0 -p 8083 -f -n -a test1.js -a one -a two -a three
Downloading.. https://storage.googleapis.com/nanos/release/nightly/nanos-nightly-linux.tar.gz
booting /home/am11/.ops/images/test1.js.img ...
You specified hardware acceleration, but it is not supported
Are you running inside a vm? If so disable accel with --accel=false

en1: assigned 10.0.2.15
args:  [ '/node_v11.5.0/node', '/test1.js', 'one', 'two', 'three' ]
exit status 1
```

is this by design?

fabioDMFerreira:
> It looks like ops-pkg command with current release (0.1.19) does not respect `Args` with `-c config.json`:
> 
> ```shell
> $ cat > test1.js <<<'console.log("args: ", process.argv)'
> $ cat > config.json <<<'{ "Args":["one","two","three"] }'
> 
> $ ops pkg load node_v11.5.0 -p 8083 -f -n -a test1.js -c config.json
> booting /home/am11/.ops/images/test1.js.img ...
> You specified hardware acceleration, but it is not supported
> Are you running inside a vm? If so disable accel with --accel=false
> 
> en1: assigned 10.0.2.15
> args:  [ '/node_v11.5.0/node', '/test1.js' ]
> exit status 1
> ```
> 
> on the other hand, the command-line args work:
> 
> ```shell
> $ ops pkg load node_v11.5.0 -p 8083 -f -n -a test1.js -a one -a two -a three
> Downloading.. https://storage.googleapis.com/nanos/release/nightly/nanos-nightly-linux.tar.gz
>  100% |████████████████████████████████████████|  [0s:0s]
> booting /home/am11/.ops/images/test1.js.img ...
> You specified hardware acceleration, but it is not supported
> Are you running inside a vm? If so disable accel with --accel=false
> 
> en1: assigned 10.0.2.15
> args:  [ '/node_v11.5.0/node', '/test1.js', 'one', 'two', 'three' ]
> exit status 1
> ```
> 
> is this by design?

Not sure how this should work if we have arguments on command flags and on the configuration file. My first impulse is to say those arguments should be concatenated. However, other configuration properties are overridden if the arguments are specified on command flags.

Any thoughts about this? @eyberg @am11

am11:
@fabioDMFerreira, the problem described in https://github.com/nanovms/ops/pull/947#issuecomment-795354646 is that `ops pkg`, unlike other commands, disregards `Args` in `config.json` when building the image. Should it be changed? This PR hasn't touched ops-pkg because of this.

Regarding override logic (which applies to all commands), I think current approach is a good tradeoff, i.e. command line arguments (`-a foo`) will take precedence over `Args` in config.json as a whole. Ideally we can take the "union" of arguments, but that would require us to implement all kinds of argument parsing behaviors. e.g.  `ops run myApp -a -p:MyParameter="foo ; \"bar\", 'baz'"`.. We can open a separate issue to discuss it in more details..

fabioDMFerreira:
This change might affect `ops pkg`. If it keeps running smoothly PR is ok

am11:
As far as I can tell it has no effect on ops-pkg whatsoever.. if you have a repro please let me know.

am11:
Opened #952 for `ops pkg -c confgWithArgs.json` vs. `ops pkg -a foo` behavior mismatch.

PR #969, opened 2021-03-15
Remove default gateway/netmask handling from ops

am11:
Defaults are now set by nanos, so it works consistently across `ops run`, `ops instance create` and other similar commands.

### DIFF
--- a/cmd/flags_run_local_instance.go
+++ b/cmd/flags_run_local_instance.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"debug/elf"
 	"fmt"
-	"net"
 	"strconv"
 
 	"github.com/nanovms/ops/types"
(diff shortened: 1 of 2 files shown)


cmd/flags_run_local_instance.go
```diff
@@ -83,44 +82,27 @@ func (flags *RunLocalInstanceCommandFlags) MergeToConfig(c *types.Config) (err e
 		c.RunConfig.BridgeName = flags.BridgeName
 	}
 
-	c.RunConfig.Verbose = flags.Verbose
-	c.RunConfig.Bridged = flags.Bridged
-	c.RunConfig.Accel = flags.Accel
-	c.Force = flags.Force
-
-	if flags.IPAddress != "" && isIPAddressValid(flags.IPAddress) {
+	if flags.IPAddress != "" {
```
fabioDMFerreira, line 85: Probably nanos discards invalid IP addresses, but I don't see a reason why not to verify whether the IP address is valid in ops side. It would be good to show some warning messages in the cases that IP address, network mask and default gateway are invalid, so user can fix it.

fabioDMFerreira:
We're supporting old nanos versions through the command flag `--nanos-version`, which don't set the defaults. Depending on nanos version used, we can remove the ops default values.

cmd/flags_run_local_instance.go
am11: nanos validates the IP address.

fabioDMFerreira: As I said, even having nanos validating IPs I don't see a reason why not having ops helping users identifying invalid IPs.

am11:
> We're supporting old nanos versions through the command flag --nanos-version.

There is no ops version released with default values, so it is a perfect time for consolidation. and besides we  haven't reached v1x, so breaking change isn't a concern.

fabioDMFerreira:
We have nanos versions released that don't set the defaults.

cmd/flags_run_local_instance.go
am11: This is a repetition and it only affects ops-run, while nanos implementation affects all commands consistently.

am11:
As said, we haven't reached v1x, so breaking change isn't a concern. but lets run it through:

`--nanos-version` and default values both are unreleased at the moment.
so once ops with `--nanos-version` option is released; if user has only specified IPAddress (without NetMask or Gateway) && user is has set an older `--nanos-version` (0.1.32 or below) && user is using anything other than `ops run` command, then before they get the DHCP resolved address and after it is a static address (which is mostly un-observable by the user)..

1. how likely it is to happen?
2. how likely it is going to be a problem?
3. what is the guarantee `--nanos-version`  is giving on old version compatibility anyway? e.g. what happens if we use `ops --nanos-version 0.1.0 -t hyper-v` which is comparatively a new addition.


cc @eyberg, @francescolavra

eyberg:
the --nanos-version flag was implemented for folks that want to lock in a known 'good' version of nanos, ops being more of a build/orchestration tool shouldn't conflict too much, for example much of the test suite in nanos doesn't actually use ops - it's just make/qemu

fabioDMFerreira:
As we're setting the defaults on Nanos, I don't see a reason to keep setting the defaults in ops. 

However, in this PR we're removing important verifications that prevent users from setting invalid values on IP/netmask/gateway. This way a user would be able to set random strings in these configuration properties that might throw unexpected errors while creating images.

am11:
No this PR is moving validation from ops to nanos.

fabioDMFerreira:
Without the isIPAddressValid helper, nothing prevents ops from setting IP addresses like `cheese` or `tomato`. Not sure what happens on nanos if we set these values in the image manifest file.

am11:
The helper does not help `ops instance create` on ops side and only favors `ops run` command. By moving the validation to nanos (we have implemented it in nanos with better and more robust gateway resolution), this ensures that all commands get the equal treatment pertaining to validation and setting the default values.

fabioDMFerreira:
I think we're talking about two different issues.

Issue 1: It's ok to remove defaults in ops because we're setting them on nanos.

Issue 2 (https://github.com/nanovms/ops/issues/979): We can set whatever ip/gateway/netmask value in the configuration file, so at least we have to guarantee they are valid IP addresses. 

Merging this PR because solves issue 1.

am11:
nanos does both: performs validation and set defaults based on the validation results; during the runtime.

If issue2 is about validating IP address and throwing error if IP address was invalid, that will definitely improve the user-experience. If the validation is silently correcting the value (setting to default or empty string), then it is potentially duplicating what nanos does.

Another issue, which was one of the motivation of this PR, is that we should try to bring about functional equivalence by de-duplication of code. This is to make the unique responsibilities of each command self-explanatory (ops-run, ops-deploy, ops-build and ops-instance-create). e.g. going by principle of least surprise:
> `ops-build` followed by `ops-run`

should carry out similar validations as:

> `ops-build` followed by `ops-instance-create`

Issue #577, opened 2020-08-11
oci (oracle) cloud target


eyberg:
no one has asked for this but prob. is super straight forward - i'd imagine built on kvm

eyberg:
i didn't know that vbox was tied into oci but that makes total sense https://github.com/nanovms/ops/issues/888

eyberg:
initial support is now in for this

PR #970, opened 2021-03-18
add oci support

fabioDMFerreira:
#577


oci/oci.go
```diff
@@ -0,0 +1,148 @@
...
+	domains, err := identityClient.ListAvailabilityDomains(context.TODO(), identity.ListAvailabilityDomainsRequest{CompartmentId: &p.compartmentID})
+	if err != nil {
+		return
+	}
+
+	if len(domains.Items) > 0 {
+		p.availabilityDomain = *domains.Items[0].Name
+	}
+
+	// instanceID := "ocid1.instance.oc1.us-sanjose-1.anzwuljrltn4jiicpxtrukrjibwzvpxvhjcjvvt3kmxc3civ24vbmv3iskdq"
```
eyberg, line 135: can we remove all of this?

oci/oci_image.go
```diff
@@ -0,0 +1,278 @@
...
+		return "", err
+	}
+
+	return imagePath, nil
+}
+
+func (p *Provider) createQcow2Image(c *types.Config) (imagePath string, err error) {
+	qcow2ImagesDir, err := findOrCreateqcow2ImagesDir()
+	if err != nil {
+		return
```
eyberg, line 49: return err?

oci/oci_image.go
```diff
@@ -0,0 +1,278 @@
...
+	args := []string{
+		"convert",
+		"-O", "qcow2",
+		c.RunConfig.Imagename, imagePath,
+	}
+
+	cmd := exec.Command("qemu-img", args...)
+	_, err = cmd.CombinedOutput()
+	if err != nil {
+		return
```
eyberg, line 63: same

oci/oci_image.go
```diff
@@ -0,0 +1,278 @@
...
+func (p *Provider) DeleteImage(ctx *lepton.Context, imagename string) (err error) {
+
+	image, err := p.getImageByName(ctx, imagename)
+	if err != nil {
+		return
+	}
+
+	_, err = p.computeClient.DeleteImage(context.TODO(), core.DeleteImageRequest{ImageId: &image.ID})
+
+	return
```
eyberg, line 247: return err?

eyberg (approved):

oci/oci.go
fabioDMFerreira: Thanks for this. Playground code

oci/oci_image.go
fabioDMFerreira: Yes. If we initialize the variables on return arguments (`(imagePath string, err error)`) we don't have to specify values on return.

fabioDMFerreira: so yes. The error is returned.

eyberg: ah i see it's named ok

PR #976, opened 2021-03-22
Fix ENA detection when creating AWS AMI.

maxi-k:
I could not get ops to create AMIs that work with AWS Elastic Network Adapter instances.
This should work after  PR #871 was merged.
The reported error with a `--flavor c5n.xlarge` flag, for example, was:
```
InvalidParameterCombination: Enhanced networking with the Elastic Network Adapter (ENA) 
is required for the 'c5n.xlarge' instance type. 
Ensure that you are using an AMI that is enabled for ENA.
	status code: 400, request id: xxxx-xxxx-xxxx
```

It turns out that the ENA-support detection based on the instance type (flavor) was not working for most instance types because the `NitroInstanceTypes` array contained capitalized instance names, but the search function searched for uppercase instance names (e.g `C5n` vs `C5N`).

This PR fixes this by making everything lowercase.
I also converted the `NitroInstanceTypes` array to a map to replace the search with a simple map lookup.
This is not necessary to fix the bug but results in nicer code imho.

I tested this with `c5n` instances and it works.


fabioDMFerreira:
@maxi-k thanks for this

PR #981, opened 2021-04-04
Hyper-V/Azure: add support for UEFI-based Generation-2 instances

francescolavra:
When creating an image for Hyper-V or Azure, the `--type` command line flag (or alternatively, the "ImageType" attribute under "CloudConfig" in the configuration file) can be used to specify a Generation-2 image type, that can run on Generation-2 instances (as well as Generation-1 instances, except for some types of Azure VMs).
In Azure, the VM generation when creating an instance is chosen automatically based on the image type; the default instance flavor has been changed from A1V2 to B1s, because A1V2 instances don't support Generation-2 images.
To specify the VM generation when creating a Hyper-V instance, the `-f` command line flag (or alternatively, the Flavor attribute under "CloudConfig" in the configuration file) can be used.
Configuration file example for Hyper-V:
```
{
    "CloudConfig": {
        "ImageType":"gen2",
        "Flavor": "gen2"
    }
}
```
Images created with the "gen2" option have an additional partition containing the UEFI bootloader file (which will be part of future Nanos releases).


am11:
For gen1 image, we can only create gen1 instance; for gen2, we can create gen2 and gen1 instances.

if we can skip or rename `Flavor` (which sounds a bit too generic) to accompany `ImageType`, that would be cool. e.g. `"DeployAsGen2": true` with cmd flag `--deploy-as-gen2`
`"PreferGen2": true` with cmd flag `--prefer-gen2`
`"ImageTypeForDeployment": "gen2"` with cmd arg `--deployment-type gen2`

fabioDMFerreira (approved):

fabioDMFerreira:
@am11 flags names must be generic because this library supports other providers besides Hyper-v and Azure.

am11:
@fabioDMFerreira, not sure what you mean, but it is:

`ImageType: gen1`  and `Flavor: gen1`
vs.
`ImageType: gen1` and `ImageTypeForDeployment: gen1`

I am not seeing we are improving or regressing anything in terms of generic here..

francescolavra:
@fabioDMFerreira was referring to command flags such as `--deploy-as-gen1` and `--prefer-gen1`, which are not generic.
As for adding a new configuration option such as "ImageTypeForDeployment", since we already have "Flavor" (which is used for a number of cloud providers), having an additional option that does the same as what "Flavor" can do seems unneeded and may cause confusion.

am11:
Ah, I didn't notice Flavor is an existing option. Makes sense to keep it as is.

PR #993, opened 2021-04-08
de-register image from virtualbox on deleting image

### LINKED ISSUE #989: deleting vbox images should issue a 'closemedium' with --delete flag
I noticed after deleting an image from vbox the image stored in ~/.ops/vdi_images was gone but it was still showing up in ``` VBoxManage list hdds```

I had to manually issue ```VBoxManage closemedium disk 5ba50fc4-415e-4eb3-9694-d592ac71e465 --delete``` otherwise when trying to spin up an instance it would complain (not sure if it would complain if i gave it a different name or not?)

fabioDMFerreira:
(closes #989)

### DIFF
--- a/vbox/vbox_image.go
+++ b/vbox/vbox_image.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/nanovms/ops/types"
+	"github.com/nanovms/ops/wsl"
 
 	"github.com/nanovms/ops/lepton"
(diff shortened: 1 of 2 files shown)


fabioDMFerreira:
I tried to delete the image and created a new one with the same name and launch an instance with no issues.

eyberg (approved):

Issue #789, opened 2020-11-24
local package should complain if given an invalid elf


eyberg:
we have elf validity checks for ops run but I accidently copied a mach-o for a local pkg and notice we don't do a check there so you get a nasty recursive trace

arknable:
@eyberg I will check this.

arknable:
Tried to reproduce using mac os node binary
```
en1: assigned 10.0.2.15
signal 11 received by tid 1, errno 0, code 1
   fault address 0x818085
   core dump (unimplemented)
exit status 255

```

arknable:
Following is the message output if the package's executable is invalid in linux:
```
package node_v14.2.0 has invalid executable
```

eyberg:
yeh - to clarify it would be good to complain if 'ops run' is given a mach-o as well

PR #1024, opened 2021-05-09
Validates package executable in Linux

arknable:
Validates package executable in Linux before the package loaded.


cmd/cmd_pkg.go
```diff
@@ -230,6 +229,24 @@ func loadCommandHandler(cmd *cobra.Command, args []string) {
 		exitWithError(err.Error())
 	}
 
+	// Validate local package
+	packageFolder := filepath.Base(pkgFlags.PackagePath())
```
eyberg, line 233: we're really the only ones generating packages today so while this is ok we're mainly concerned with ensuring that one isn't trying to run a macho-o

arknable:
@eyberg it actually works with mach-o
```
Downloading.. https://storage.googleapis.com/nanos/release/nightly/nanos-nightly-linux.tar.gz
package node_v14.2.0 has invalid executable

```

I used [this mach-o binary](https://github.com/JonathanSalwan/binary-samples/blob/master/MachO-OSX-x64-ls) for the test.

fabioDMFerreira:
Commands like `ops run`, `ops image create` and `ops build` also require an elf file to run, so it would be useful to have this validation there. I know #789 is only for validating the elf on local packages though.

To avoid having duplicated code we could place the validation in `cmd/flags_build_image.go`.

cmd/cmd_pkg.go
```diff
@@ -230,6 +229,24 @@ func loadCommandHandler(cmd *cobra.Command, args []string) {
...
+	executableName := c.Program
+	if strings.Contains(executableName, packageFolder) {
+		executableName = filepath.Base(executableName)
+	}
+	executablePath := filepath.Join(pkgFlags.PackagePath(), executableName)
+
+	if runtime.GOOS == "linux" {
```
fabioDMFerreira, line 240: Does it make sense to do this for `darwin`?

eyberg: yeh - no need to ensure a platform chk here - the real problem is when someone is building on a mac and forgets to target for linux

arknable:
`run`, `image create`, and `build` already have executable checking and raises following error
`only ELF binaries are supported. Is thia a Linux binary? run "file /home/arknable/.ops/packages/node_v14.2.0/node" on it`

So I will just refactor the validation code as a separated function and call it upon package load

fabioDMFerreira:
@arknable I know what you mean. It's strange that the program is being validated on the method `getSharedLibs`. Anyway, it's a good idea to have the validation code in a dedicated function.

Just make sure you avoid platform-specific code (` runtime.GOOS == "linux"`). Use instead the files `ldd_darwin` and `ldd_linux`. It's ok to create a stub for darwin to make it compile as there is no `isElf` method.

arknable:
> @arknable I know what you mean. It's strange that the program is being validated on the method `getSharedLibs`. Anyway, it's a good idea to have the validation code in a dedicated function.
> 
> Just make sure you avoid platform-specific code (` runtime.GOOS == "linux"`). Use instead the files `ldd_darwin` and `ldd_linux`. It's ok to create a stub for darwin to make it compile as there is no `isElf` method.

@fabioDMFerreira not sure I understand correctly. as we are also checking elf in darwin, I think `isELF` and the validation function should be put inside `ldd.go` so they are available on both os, right ?

fabioDMFerreira:
In the current master branch the `isElf` only exists in `ldd_linux`. I thought the elf validation in darwin had to be done in a different way.

Could you rebase your branch to solve the conflicts? Thanks

arknable:
> In the current master branch the `isElf` only exists in `ldd_linux`. I thought the elf validation in darwin had to be done in a different way.

ah ok, I understand

arknable:
> Could you rebase your branch to solve the conflicts? Thanks

yes I dont know what is the problem but there is no conflict in that file, I already rebased the file against master but the old code keep showing here. I am trying to use github UI to resolve the conflict.

arknable:
> yes I dont know what is the problem but there is no conflict in that file, I already rebased the file against master but the old code keep showing here. I am trying to use github UI to resolve the conflict.

ok it works

fabioDMFerreira:
I've also faced this issue before. I recommend creating a new branch and cherry-pick the commits you want to push.

cmd/cmd_pkg.go
```diff
@@ -228,6 +226,14 @@ func loadCommandHandler(cmd *cobra.Command, args []string) {
 		exitWithError(err.Error())
 	}
 
+	// Validate executable
```
fabioDMFerreira, line 229: I think we don't need this comment. The code is pretty clear.

arknable:
> I've also faced this issue before. I recommend creating a new branch and cherry-pick the commits you want to push.

yeah you're right the  web editor merged master to PR's branch. I will move the commits to new branch

fabioDMFerreira:
The build in circleci is failing. If you want to debug locally you can run the script `hack/verify-golint.sh`.

arknable:
> The build in circleci is failing. If you want to debug locally you can run the script `hack/verify-golint.sh`.

ah ok, thanks will do that instead.

arknable:
This PR replaced by https://github.com/nanovms/ops/pull/1032

PR #1033, opened 2021-05-14
Fix Port Conflict Error

arknable:
Fixed port conflict error message to be human-readable

### DIFF
--- a/cmd/cmd_pkg.go
+++ b/cmd/cmd_pkg.go
@@ -3,17 +3,12 @@ package cmd
 import (
 	"bufio"
 	"fmt"
-	"log"
-	"net"
 	"os"
-	"os/exec"
 	"path"
 	"path/filepath"
 	"regexp"
-	"runtime"
 	"sort"
 	"strings"
-	"time"
 
 	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


cmd/cmd_root.go
```diff
@@ -31,6 +31,27 @@ func GetRootCommand() *cobra.Command {
 				}
 			}
 
+			showErrors, err := cmd.Flags().GetBool("show-errors")
```
fabioDMFerreira, line 34: isn't this being handled by `globalFlags.MergeToConfig(config)`?

cmd/cmd_pkg.go
```diff
@@ -246,3 +245,14 @@ func loadCommandHandler(cmd *cobra.Command, args []string) {
 		exitWithError(err.Error())
 	}
 }
+
+func excludeWhitespaces(arr []string) []string {
```
fabioDMFerreira, line 249: does it make sense to move to another file?

fabioDMFerreira (changes requested):

cmd/cmd_root.go
arknable: @fabioDMFerreira oh you right ... I will move that line outside the `if` block

arknable: Fixed by https://github.com/nanovms/ops/pull/1033/commits/70a813bb5130c1f9011f2d055f004716c5e21bf3

cmd/cmd_pkg.go
arknable: @fabioDMFerreira how about moved to separated `util` package ?

arknable: Fixed by https://github.com/nanovms/ops/pull/1033/commits/ca51410eb0f979e998ebac0d65c60e2b105790e3

fabioDMFerreira (approved):

PR #1046, opened 2021-05-19
Added bash completion script

arknable:
Added bash autocomplete script for ops commands

### DIFF
--- a/bash_completion.sh
+++ b/bash_completion.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+__nanovms_ops_completions() {
+    if [ "${#COMP_WORDS[@]}" = "2" ]; then
+        COMPREPLY+=($(compgen -W "help build deploy image instance pkg profile run update version volume" "${COMP_WORDS[1]}"))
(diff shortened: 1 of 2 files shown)


arknable:
@eyberg is there any document about installation package which being put in the cloud storage ? I need to put the bash script into that package.

fabioDMFerreira:
it is in this repo https://github.com/nanovms/ops/blob/master/install.sh

arknable:
@fabioDMFerreira oh I just noticed this code
```
if [ "$ARCH" = "aarch64" ]; then
    BINARY_URL="$RELEASES_URL/${OS}/${ARCH}/ops"
  else 
    BINARY_URL="$RELEASES_URL/${OS}/ops"
  fi
  ```
  
  I thought we download somekind of zipped package. so which script that put `ops` binary to cloud ? I need to put `bash_completion.sh` script to the cloud also.

arknable:
@eyberg Completion script added to installation process.

eyberg (approved):

Issue #586, opened 2020-08-20
esx multiple volume mounting


eyberg:

eyberg:
esx instructions now available at https://github.com/nanovms/nanos/issues/1249

eyberg:
https://github.com/nanovms/nanos/pull/1272

fabioDMFerreira:
Blocked by next issue
https://github.com/vmware/govmomi/issues/2174

Although we can easily create virtual disks on datastorage and mount them with virtual machines, the process of getting information about the existing volumes is complex without the usage of the volume manager provided by `https://github.com/vmware/govmomi`.

Useful links:
- Object Manager documentation (https://pkg.go.dev/github.com/vmware/govmomi@v0.22.2/vslm)

PR #1049, opened 2021-05-20
Add vsphere volume operations (#586)

fabioDMFerreira:
* Power on a vsphere instance after creating it.

TODO:
* Deleting vsphere volume
* Using vsphere object manager on volume operations

### DIFF
--- a/vsphere/vsphere_volume.go
+++ b/vsphere/vsphere_volume.go
@@ -55,15 +55,12 @@ func (v *Vsphere) CreateVolume(ctx *lepton.Context, name, data, size, provider s
 		return vol, err
 	}
 
-	// TODO: Implement an alternative solution to register the disk using the object manager.
-	// The advantages are the simplification of operations like getting attachments and deleting
-	// I could not having this working due to the error stated in https://github.com/vmware/govmomi/issues/2174.
-	// objectManager := vslm.NewObjectManager(ds.Client())
+	objectManager := vslm.NewObjectManager(ds.Client())
 
-	// _, err = objectManager.RegisterDisk(context.TODO(), ds.NewURL("volumes/"+vol.Name+".vmdk").String(), vol.Name)
-	// if err != nil {
-	// return vol, fmt.Errorf("register disk: %v", err)
-	// }
+	_, err = objectManager.RegisterDisk(context.TODO(), ds.NewURL("volumes/"+vol.Name+".vmdk").String(), vol.Name)
+	if err != nil {
+		return vol, fmt.Errorf("register disk: %v", err)
+	}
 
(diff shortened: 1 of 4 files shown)


vsphere/vsphere_volume.go
```diff
@@ -1,29 +1,295 @@
...
+	if err != nil {
+		return vol, err
+	}
+
+	// TODO: Implement an alternative solution to register the disk using the object manager.
+	// The advantages are the simplification of operations like getting attachments and deleting
+	// I could not having this working due to the error stated in https://github.com/vmware/govmomi/issues/2174.
+	// objectManager := vslm.NewObjectManager(ds.Client())
+
+	// _, err = objectManager.RegisterDisk(context.TODO(), ds.NewURL("volumes/"+vol.Name+".vmdk").String(), vol.Name)
```
eyberg, line 63: can we cut?

vsphere/vsphere_volume.go
```diff
@@ -1,29 +1,295 @@
...
+
+	for _, file := range files {
+		path := file.GetFileInfo().Path
+		if !strings.Contains(path, "-flat.vmdk") {
+			*vols = append(*vols, lepton.NanosVolume{Name: path})
+		}
+	}
+
+	// TODO: use object manager to get all volumes listed and convert them to nanos volumes, blocked by https://github.com/vmware/govmomi/issues/2174
+	// disks, err := v.getAllVolumes(ds)
```
eyberg, line 135: minus the todo; can we cut?

vsphere/vsphere_volume.go
```diff
@@ -1,29 +1,295 @@
...
+	// }
+
+	return vols, nil
 }
 
 // DeleteVolume is a stub to satisfy VolumeService interface
 func (v *Vsphere) DeleteVolume(ctx *lepton.Context, name string) error {
-	return nil
+	// TODO: Blocked by https://github.com/vmware/govmomi/issues/2174
+	// f := find.NewFinder(v.client, true)
```
eyberg, line 146: this chunk of code should be rm'd

eyberg (approved):
just need to get rid of all the commented out code

fabioDMFerreira:
Once we overpass the problem in https://github.com/vmware/govmomi/issues/2174, we can uncomment the code and remove the current implementation.

The commented code is the code using the alternative solution. Basically, the code just need to be tested.

eyberg:
I would either leave the code in or get rid of it

fabioDMFerreira:
In that case, I'll opt to uncomment the code and delete the current implementation. It's unlikely it will work at the first attempt, but it's a cleaner solution.

PR #1065, opened 2021-06-05
Always display error messages

arknable:
Skip error flag check so error messages always displayed

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -47,30 +47,30 @@ func (l *Logger) Log(message string, a ...interface{}) {
 // Info checks info level is activated to write the message
 func (l *Logger) Info(message string, a ...interface{}) {
 	if l.info == true {
-		l.Log(ConsoleColors.Blue()+message+ConsoleColors.White(), a...)
+		l.Log(ConsoleColors.Blue()+message+ConsoleColors.Reset(), a...)
 	}
 }
 
 // Warn checks warn level is activated to write the message
 func (l *Logger) Warn(message string, a ...interface{}) {
 	if l.warn == true {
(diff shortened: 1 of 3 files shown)


log/logger.go
```diff
@@ -60,16 +60,12 @@ func (l *Logger) Warn(message string, a ...interface{}) {
 
 // Errorf checks error level is activated to write the formatted message
 func (l *Logger) Errorf(message string, a ...interface{}) {
-	if l.err == true {
-		l.Log(ConsoleColors.Red()+message+ConsoleColors.White(), a...)
-	}
+	l.Log(ConsoleColors.Red()+message+ConsoleColors.White(), a...)
```
francescolavra, line 63: No related specifically to this change, but setting the console color to white after the error message (the same applies to warning, info and debug messages as well) should be replaced by restoring the default color (there is a specific ANSI color sequence for that, which is "\033[0m"), otherwise if someone uses a terminal with a default foreground color other than white, after running ops the console text color will change and in extreme cases (if the background color is also white) the terminal may not be usable anymore.

eyberg:
1062 is better;
1054 is better,
1055 is better

thanks for fixing

eyberg:
i'd just fix @francescolavra's comment and then we could merge and close these others out

eyberg (approved):

log/logger.go
arknable: Fixed by https://github.com/nanovms/ops/pull/1065/commits/2059dffce642578e685ac1e196bc13571ae9a91d. I have tested the changes in my terminal with various dark and light presets ... looks fine.

francescolavra (commented):
Looks good, thanks

PR #1068, opened 2021-06-06
Override PackageBaseURL and PackageManifestURL

arknable:
Added support to override `PackageBaseURL` and `PackageManifestURL`. The override locations based on priority in order are `$HOME/.opsrc` then environment variable `OPS_PACKAGE_BASE_URL` and `OPS_PACKAGE_MANIFEST_URL` .... meaning environment variable will override `.opsrc`.

Added also support for local folder using `file://` prefix, e.g. if we have repo directory in local filesystem and we want to use it, we can override it like this
```
{
	"PackageBaseURL": "file:///home/arknable/Development/Project/nanovms/example/package_repo",
	"PackageManifestURL": "file:///home/arknable/Development/Project/nanovms/example/package_repo/manifest.json"
}
```

or using environment variables with above values.

### DIFF
--- a/lepton/package.go
+++ b/lepton/package.go
@@ -70,7 +70,7 @@ func DownloadPackage(name string, config *types.Config) (string, error) {
 
 	// Check environment variable override
 	ePkgBaseURL := os.Getenv("OPS_PACKAGE_BASE_URL")
-	if len(ePkgBaseURL) > 0 {
+	if ePkgBaseURL != "" {
 		pkgBaseURL = ePkgBaseURL
 	}
 
@@ -137,7 +137,7 @@ func GetPackageList(config *types.Config) (*map[string]Package, error) {
 
 	// Check environment var override
 	ePkgManifestURL := os.Getenv("OPS_PACKAGE_MANIFEST_URL")
-	if len(ePkgManifestURL) > 0 {
+	if ePkgManifestURL != "" {
 		pkgManifestURL = ePkgManifestURL
 	}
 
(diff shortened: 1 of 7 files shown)


lepton/package.go
```diff
@@ -49,25 +49,120 @@ func DownloadPackage(name string) (string, error) {
...
+
+	destFile, err := os.Create(packagepath)
+	if err != nil {
+		return "", err
+	}
+	defer destFile.Close()
+
+	progressCounter := NewWriteCounter(int(srcStat.Size()))
+	progressCounter.Start()
+	_, err = io.Copy(destFile, io.TeeReader(srcFile, progressCounter))
```
eyberg, line 115: should return this error if exists

eyberg (approved):

lepton/package.go
arknable: Fixed by https://github.com/nanovms/ops/pull/1068/commits/17ca5223709c4f1fe1da31ac1e96d534fb207b44

arknable:
I have rebased the branch but still detected as conflicted. I will just create new branch with the same name and cherry pick from the old one.

arknable:
@eyberg Rebased, please recheck

lepton/package.go
```diff
@@ -49,25 +49,123 @@ func DownloadPackage(name string) (string, error) {
...
+	if config != nil {
+		cPkgBaseURL := strings.Trim(config.PackageBaseURL, " ")
+		if len(cPkgBaseURL) > 0 {
+			pkgBaseURL = cPkgBaseURL
+		}
+	}
+
+	// Check environment variable override
+	ePkgBaseURL := os.Getenv("OPS_PACKAGE_BASE_URL")
+	if len(ePkgBaseURL) > 0 {
```
eyberg, line 73: not a big deal, but usually the idiom ePkgBaseURL != "" is typically preferred

eyberg (approved):

lepton/package.go
arknable: ok, fixed by https://github.com/nanovms/ops/pull/1068/commits/df339a12c9f4a0dd3f31453d1ad63413883ebf6d

PR #1069, opened 2021-06-07
Allow `Language`, `Runtime` and `Description` properties in manifest configuration

quantumsheep:
Whenever a package is added to `local_packages`, the properties `Language`, `Runtime` and `Description` are removed. Making `ops pkg list --local` not displaying those information even though having columns dedicated for them.

This PR fixes this issue by keeping the related properties.

### DIFF
--- a/cmd/flags_pkg.go
+++ b/cmd/flags_pkg.go
@@ -59,6 +59,10 @@ func (flags *PkgCommandFlags) MergeToConfig(c *types.Config) (err error) {
 	c.Program = pkgConfig.Program
 	c.Version = pkgConfig.Version
 
+	c.Language = pkgConfig.Language
+	c.Runtime = pkgConfig.Runtime
+	c.Description = pkgConfig.Description
+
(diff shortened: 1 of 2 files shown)


eyberg (changes requested):
Thanks!

Is it possible to refactor this to use the struct found in lepton/Package?

quantumsheep:
> Thanks!
> 
> Is it possible to refactor this to use the struct found in lepton/Package?

I'm not sure about what you want to do? Extending `Config` by using `Package`?

eyberg:
ah, i see what your issues, public packages are defined in manifest.json and map to the struct here -> https://github.com/nanovms/ops/blob/master/lepton/package.go#L31

however the local packages doesn't have a manifest with them as that's un-defined currently so you're stuffing them in the package.manifest and pulling it from there

the config prob isn't the best place for that because it is mostly used for running instances, but I suppose that's fine for now

eyberg (approved):

PR #1070, opened 2021-06-08
Implement a command to create a package from an archive or a folder

quantumsheep:
This PR introduces the command `ops pkg add`. Giving it a folder or a `tar.gz` archive will copy it to `local_packages`.

A flag `--name`/`-n` is available to choose the package's name. If not given it will generate a random 16 characters identifier (kinda like Docker containers name).

### Example
```
$ tree --noreport node_v16.3.0 
node_v16.3.0
├── node
├── package.manifest
└── sysroot
    ├── lib
    │   └── x86_64-linux-gnu
    │       ├── libc.so.6
    │       ├── libdl.so.2
    │       ├── libgcc_s.so.1
    │       ├── libm.so.6
    │       └── libpthread.so.0
    ├── lib64
    │   └── ld-linux-x86-64.so.2
    └── usr
        └── lib
            └── x86_64-linux-gnu
                └── libstdc++.so.6

$ cd node_v16.3.0
$ tar -czvf package.tar.gz *
$ ops pkg add package.zip --name node_v16.3.0
node_v16.3.0
$ ops pkg load -l node_v16.3.0 -a app.js
booting /Users/qtmsheep/.ops/images/node ...
en1: assigned 10.0.2.15
Hello, World!
exit status 1
```


eyberg:
we don't support anything other than .tar.gz atm - is there a good reason for supporting .zip/.rar/etc?

this introduces a few dependencies for dealing with other formats and for doing a simple cp?

quantumsheep:
> we don't support anything other than .tar.gz atm - is there a good reason for supporting .zip/.rar/etc?
Mostly for portability, I really wanted to add zip file support to this command. `rar` came with the package I used for handling archives. I think if that's an issue we could only support `.tar.gz` bu why not supporting other algorithms?

> this introduces a few dependencies for dealing with other formats and for doing a simple cp?
That was the easiest way I found to copy every files of a folder without copying the folder itself. I'm pretty new to Go, is there another better solution?

eyberg:
there's not really a problem - it's more of a we don't want to add code that doesn't have a use-case that's all; ideally we'd standardize one one format and optimize for compression speed/ratio; having said that, we're not against it if you have a usecase that would require it

can you describe your zip usecase a bit?

as for the cp - i understand you want to copy the entire folder, however, packages are distributed already tarr'd/zip'd so it sounds like that's more of an ease-of-use for working locally?

quantumsheep:
> can you describe your zip usecase a bit?

I'm using macOS, zipping is the easiest format to compress to because it's natively implemented. I'm thinking about Windows where it's also the default format.

> as for the cp - i understand you want to copy the entire folder, however, packages are distributed already tarr'd/zip'd so it sounds like that's more of an ease-of-use for working locally?

I copy the content because we want to keep the original directory where it actually is (for development purpose). For archives it's because `os.Rename` didn't seem to work when used over a directory located in temp.

eyberg:
got it, yeh I think for now we probably won't want to support more than one format; adding more formats will force other registries and tooling to support those formats as well; and down the road we might want to look at optimizing how we transfer them around too (eg: force use of one method); when we add more features to the 'public' registry for instance that will probably want us to look at methods of saving on bandwidth egress for instance

ops only runs on WSL for windows; osx && linux so tar is already installed and uses the go implementation anyways

as for the copy folder - totally ok with having that functionality - we'd just prefer not including other modules if not necessary;

would it be possible to remove those external deps in favor of something much simpler such as:

https://blog.depa.do/post/copy-files-and-directories-in-go#toc_2
https://stackoverflow.com/questions/51779243/copy-a-folder-in-go

if you can clean that up we can take another look and merge

quantumsheep:
@eyberg Seems good

eyberg (approved):

eyberg:
thanks!

PR #1075, opened 2021-06-09
adding in manifest passthrough for pkgs

eyberg:


lepton/image.go
```diff
@@ -223,6 +223,10 @@ func BuildPackageManifest(packagepath string, c *types.Config) (*fs.Manifest, er
 		}
 	}
 
+	for k, v := range c.ManifestPassthrough {
+		m.AddPassthrough(k, v)
+	}
```
francescolavra, line 228: LGTM. Alternatively, these lines can be moved from BuildManifest() to setManifestFromConfig(), which is called by both  BuildManifest() and BuildPackageManifest().

PR #1078, opened 2021-06-11
Added option to override home path via OPS_HOME environment variable

arknable:
`OPS_HOME` value should be the parent directory in which `.ops` directory located, hence the home directory will be `$OPS_HOME/.ops`. if `$OPS_HOME/.ops` not exists, it will be created automatically.

### DIFF
--- a/cmd/cmd_pkg.go
+++ b/cmd/cmd_pkg.go
@@ -280,6 +280,8 @@ func loadCommandHandler(cmd *cobra.Command, args []string) {
 	executableName := c.Program
 	if strings.Contains(executableName, packageFolder) {
 		executableName = filepath.Base(executableName)
+	} else {
+		executableName = filepath.Join(lepton.PackageSysRootFolderName, executableName)
 	}
 	lepton.ValidateELF(filepath.Join(pkgFlags.PackagePath(), executableName))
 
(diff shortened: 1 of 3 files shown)


eyberg:
this seems like it needs something more - for some commands it'll instantiate a homedir on demand; on other commands can't find the right file even though it's there

I would think if this env var is set then it should probably try and see if a proper ops home exists (is there a latest nanos installed, etc?) and if not instantiate it; for some commands it took a little bit to complete

arknable:
`ops` creates standard directories in home folder if not exists in `GetOpsHome` function.

```
arknable@arknable-x1c:~/.../ops$ ls ~/.opshome/
arknable@arknable-x1c:~/.../ops$ OPS_HOME=$HOME/.opshome ./ops image list
+------+------+------+-----------+
| NAME | PATH | SIZE | CREATEDAT |
+------+------+------+-----------+
arknable@arknable-x1c:~/.../ops$ ls ~/.opshome/
images  instances  local_packages  manifests  packages  volumes
```
So I think we need to make sure that function is being used to get ops home path.

arknable:
@eyberg could you please share which command is failing ? I only see one hardcoded usage of home dir.

arknable:
@eyberg please recheck, I have tested some commands and they seems working fine.

eyberg (approved):

PR #1079, opened 2021-06-14
Log missing files during image build instead of returning error

sanderssj:
While building an image and adding package files to it, if there is a
problem with a file such as a broken symlink, the addFile walk aborts
but the error is not handled. This has been changed so the err is logged
as an error and printed to the user but the image build continues since
missing package files aren't always required for the app to run. This fixes
some ops pkg load regressions such as in the python2.7 package.

### DIFF
--- a/lepton/image.go
+++ b/lepton/image.go
@@ -173,7 +173,7 @@ func addFilesFromPackage(packagepath string, m *fs.Manifest) {
 
 		err = m.AddFile(filePath[1], hostpath)
 		if err != nil {
-			return err
+			log.Error(err)
 		}
 		return nil
 	})
@@ -195,7 +195,7 @@ func addFilesFromPackage(packagepath string, m *fs.Manifest) {
 		vmpath := filepath.Join(string(os.PathSeparator), packageName, filePath[1])
 		err = m.AddFile(vmpath, hostpath)
 		if err != nil {
-			return err
+			log.Error(err)
 		}
 		return nil
 	})


lepton/image.go
```diff
@@ -173,7 +173,7 @@ func addFilesFromPackage(packagepath string, m *fs.Manifest) {
 
 		err = m.AddFile(filePath[1], hostpath)
 		if err != nil {
-			return err
+			log.Error(err)
```
eyberg, line 176: I think this should be log.Error(err) && then return err otherwise it hits that return nil (which could prob just be removed)

lepton/image.go
```diff
@@ -195,7 +195,7 @@ func addFilesFromPackage(packagepath string, m *fs.Manifest) {
 		vmpath := filepath.Join(string(os.PathSeparator), packageName, filePath[1])
 		err = m.AddFile(vmpath, hostpath)
 		if err != nil {
-			return err
+			log.Error(err)
```
eyberg, line 198: same

lepton/image.go
sanderssj: That would stop the walk of sysroot completely and prevent any further files from being added to the image. I think we could do that, but we should probably check addFilesFromPackage and return an error in BuildPackageManifest so the command would fail and make you fix the package. We have to return nil if we want the behavior to be what it was before, where it ignored the error.

eyberg: oh yeh, you're right

eyberg (approved):

PR #1094, opened 2021-06-22
catch json parsing issues for local packages

eyberg:


lepton/package.go
```diff
@@ -32,11 +32,11 @@ type PackageList struct {
...
 type Package struct {
-	Runtime     string `json:"runtime"`
-	Version     string `json:"version"`
-	Language    string `json:"language"`
+	Runtime     string `json:"runtime,omitempty"`
+	Version     string `json:"version,omitempty"`
+	Language    string `json:"language,omitempty"`
 	Description string `json:"description,omitempty"`
-	SHA256      string `json:"sha256"`
+	SHA256      string `json:"sha256,omitempty"`
```
francescolavra, line 39: As far as I understand, the `omitempty` directive is only used when marshaling data into JSON strings, and not when unmarshaling JSON strings, so I don't think it makes a difference here.

eyberg: you're right - it was actually just the invalid json that was blowing things up - rmd

PR #1097, opened 2021-06-25
Optimize `HomeDir()`

### LINKED ISSUE #1095: memoize homedir lookup
was looking into a sep. issue and noticed https://github.com/nanovms/ops/blob/master/lepton/homedir.go#L11 can be called a lot (for ops pkg load node)

this prob isn't that expensive of a call but the fact that it's getting called 15? times it could prob just be set after the first call as I don't think it would ever change in flight

```
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
setting homedir
home set to /Users/eyberg
```

quantumsheep:
Resolves #1095.

### DIFF
--- a/lepton/homedir.go
+++ b/lepton/homedir.go
@@ -5,18 +5,26 @@ import (
 	"os"
 )
 
+var homeDir = ""
+
 // HomeDir returns the home directory for the executing user.
 // This uses an OS-specific method for discovering the home directory.
 // An error is returned if a home directory cannot be detected.
 func HomeDir() (string, error) {
-	if home, err := os.UserHomeDir(); err == nil {
-		return home, nil
+	if homeDir != "" {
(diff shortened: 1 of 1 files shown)


eyberg (approved):
excellent - thanks!

PR #1098, opened 2021-06-27
Check qemu version in 'pkg load' and 'run' commands 

arknable:
Warn user if their installed QEMU version is older than 2.9

```
[root@localhost ops_example]# qemu-system-x86_64 --version
QEMU emulator version 1.5.3 (qemu-kvm-1.5.3-175.el7_9.4), Copyright (c) 2003-2008 Fabrice Bellard
[root@localhost ops_example]# ops run hello

supported QEMU version is 2.9 or above, please do upgrade.

root@localhost ops_example]# 
```

```
[root@localhost ops_example]# ops pkg load node_v14.2.0 -f -n -a hello.js

supported QEMU version is 2.9 or above, please do upgrade.

[root@localhost ops_example]# 
```

### DIFF
--- a/cmd/cmd_pkg.go
+++ b/cmd/cmd_pkg.go
@@ -18,6 +18,7 @@ import (
 	"github.com/nanovms/ops/lepton"
 	api "github.com/nanovms/ops/lepton"
 	"github.com/nanovms/ops/log"
+	"github.com/nanovms/ops/qemu"
 )
 
 // PackageCommands gives package related commands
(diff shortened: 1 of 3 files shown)


eyberg (changes requested):
I don't think we want this on every command - can we just move it to when it fails to run cause of a lower version instead?

arknable:
> I don't think we want this on every command - can we just move it to when it fails to run cause of a lower version instead?

ok, let me try using older qemu and see which part is failing

arknable:
@eyberg Please recheck

PR #1112, opened 2021-07-08
Other than onprem, instance's image is not a path

arknable:
For cloud-based image, check instance's image against image name instead of path.


eyberg (approved):

arknable:
Rebased

Issue #1111, opened 2021-07-06
Generate a package from a Docker image


quantumsheep:
When generating packages for `ops`, I realized that to fetch executables and their shared libraries I was depending on Docker images a lot, using `ldd` to retrieve shared libraries and then copying the files out of the container.

This use case led to the current idea - automating the fetching process of those executables and their shared libraries from Docker images.

Here's an example:
```bash
$ ops pkg from-docker node:16.3.0-alpine
```
Because we didn't give the wanted executable, we can look at the `CMD` Docker instruction inside the image's `Dockerfile`.

If there isn't `CMD` instruction or the user want to get another executable than the one used in `CMD` we could do something like that:
```bash
$ ops pkg from-docker node:16.3.0-alpine -f /usr/local/bin/node
```

The result would be a generated package (we can think of a `--name` option and more).

If this idea is accepted I would like to make it.

eyberg:
ppl have asked about this from time to time and I think looking at the cmd or passing in the name is fine; one of the differences that docker brings over what we are doing with manually created packages right now is that it provides a default filesystem which can be heavier-weight but but if the tradoff is that people don't have to guess/enumerate on what they need to include I think that's fine

rrs45:
Getting the shared libraries with the right versions has been a pain point
for me. For example, libc.so and libm.so require specific glibc versions
which may not be available on the host OS. I had to build different glibc
versions from source to generate the right libs, same case with libffi. All
open source products have an official docker image which can be used to
build ops packages, providing consistent build experience. This would be
very helpful, thanks

PR #1114, opened 2021-07-08
Add `pkg from-docker` subcommand

quantumsheep:
Relates to #1111.

There are some things in this code that might need to be discussed, especially how the shared libraries are fetched.

I needed some libraries to exchange with the Docker API:
```
github.com/docker/distribution/reference
github.com/docker/docker/api/types
github.com/docker/docker/api/types/container
github.com/docker/docker/client
github.com/docker/docker/pkg/jsonmessage
github.com/moby/term
```

It doesn't have the "automatic executable finding" that I proposed in #1111 because I couldn't find how to do it but at least we have this first iteration.

When using an image, if not present, it will be pulled automatically.

The default package name is `{repository}-{tag}`.

### Usage
```
ops pkg from-docker [image] [flags]

Flags:
  -f, --file string   target executable
  -h, --help          help for from-docker
  -n, --name string   name of the package
  -q, --quiet         quiet mode
      --verbose       verbose mode
```

### Examples
```bash
$ ops pkg from-docker node:16.3.0 -f node

$ ops pkg load -l node-16.3.0 -a a.js
booting /Users/qtmsheep/.ops/images/program ...
en1: assigned 10.0.2.15
Hello
exit status 1
```


eyberg (approved):
nifty! this is fairly seamless and will probably help a lot of people out

PR #1123, opened 2021-07-25
Skip statically linked program with pkg from-docker

### LINKED ISSUE #1115: pkg from-docker doesn't seem to like musl
```
➜  ~ ops pkg from-docker vault -f vault
latest: Pulling from library/vault
540db60ca938: Pull complete
a01b10d36ffd: Pull complete
0c6b9f118d4b: Pull complete
1c43fa2ed89c: Pull complete
bc1170290644: Pull complete
Digest: sha256:28c9a18e03194b3644f330e083431250b720ef04855de02bd6f9c06071741781
Status: Downloaded newer image for vault:latest
2021/07/08 16:14:02 Invalid library declaration: B/lib/ld-musl-x86_64.so.1: /bin/vault: Not a valid dynamic program
```

am11:
Closes #1115

### DIFF
--- a/cmd/extract_from_docker.go
+++ b/cmd/extract_from_docker.go
@@ -37,8 +37,7 @@ func ExtractFromDockerImage(imageName string, packageName string, targetExecutab
 		colors=""
 
 		read_libs() {
-			path=$(which $1)
-			ldd ${path:-$1} | rev | cut -d' ' -f2 | rev | while read lib; do
+			ldd "$1" | rev | cut -d' ' -f2 | rev | while read lib; do
 				if [ "$(echo $lib | cut -c1-1)" = "/" ]; then
 					exists=0
 					resolved_lib=$(readlink -f $lib)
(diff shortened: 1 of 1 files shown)


eyberg:
not sure why but I seem to be getting empty files on this branch when trying both  &&

```ops pkg from-docker node:16.3.0 -f node```

and

```ops pkg from-docker vault -f vault```

**Comment by am11** [comment]
Ah, nice catch! I was missing an echo. The expected format is

```
<path to program>
<list of files>
```

Currently, we do not get the errors which are emitting from the container, so even if we say `exit 1`, we get nil error from `createContainer(imageName, []string{"sh", "-c", script}, true, quiet)`. But this is something we can track separately as we would need to modify our container invocations.

eyberg:
the node example seems to work again but the vault example is still producing a 0 byte file - are you testing with a different statically linked program?

am11:
Nope I had to hoist it out of the condition. I tested and formatted in wrong order.

eyberg (approved):

eyberg:
thanks!

PR #1131, opened 2021-07-28
we are persisting versioned nanos releases but not ops - this persist…

eyberg:
…s versioned ops releases


francescolavra (commented):
Looks good to me

PR #1132, opened 2021-07-29
Commands: add support for cross-build environment

francescolavra:
This PR adds the 'env' set of commands, which allows to build an application in a Linux virtual machine and create a Nanos image from a set of files retrieved from the Linux virtual machine.
Usage summary for the new commands:
- `ops env install`: downloads a VM image from an online repository and extracts it in $OPS_HOME/crossbuild/images
- `ops env uninstall`: removes the VM image from $OPS_HOME/crossbuild/images
- `ops env build <build_file> --create-image -c config.json`: starts the build VM, connects to it via ssh, executes in sequence the set of commands specified in <build_file> (aborting the process if any of those commands returns an error), then reads any "Files", "Dirs" and "MapDirs" directives in config.json (which is a standard ops configuration file), copies the relevant files in a temporary folder in the host and uses them to create an on-prem image; finally, the build VM is stopped
- `ops env build <build_file> -r <host_path> -c config.json`: like above, but instead of copying the files in a temporary folder copies them into <host_path>, and does not create an image; the set of files copied in <host_path> can then be used to create any image the user may want (on-prem or otherwise); this is useful if the user wants to modify the set of files and/or add new files before creating an image
- `ops env copy <host_path> -c config.json`: like above, but without the build step; this is useful if the user wants to create
different images with different sets of files, and doesn't need to repeat the build step every time

### DIFF
--- a/cmd/cmd_env.go
+++ b/cmd/cmd_env.go
@@ -1,0 +1,200 @@
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/nanovms/ops/crossbuild"
+	api "github.com/nanovms/ops/lepton"
+	"github.com/nanovms/ops/log"
+	"github.com/spf13/cobra"
+)
+
+// EnvCommand provides cross-build features.
+func EnvCommand() *cobra.Command {
+	cmdInstall := &cobra.Command{
+		Use:   "install",
+		Short: "Install environment",
+		Run:   envInstall,
+	}
+
+	cmdBuild := &cobra.Command{
+		Use:   "build <build_file> [flags]",
+		Short: "Build application in environment",
+		Args:  cobra.MinimumNArgs(1),
+		Run:   envBuild,
+	}
+	flags := cmdBuild.PersistentFlags()
+	flags.BoolP("create-image", "", false, "create Nanos image from environment")
+	PersistConfigCommandFlags(flags)
+	PersistNightlyCommandFlags(flags)
+	PersistNanosVersionCommandFlags(flags)
+	PersistBuildImageCommandFlags(flags)
+
+	cmdCopy := &cobra.Command{
+		Use:   "copy <local_dir> [flags]",
+		Short: "Copy files from environment to local directory",
+		Args:  cobra.MinimumNArgs(1),
+		Run:   envCopy,
+	}
+	flags = cmdCopy.PersistentFlags()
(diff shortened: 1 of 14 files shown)


eyberg:
this currently doesn't work for me (on mac) cause it's trying to map the guest location with the same host location for some of the shared libs; fwiw this function looks off to me 
https://github.com/nanovms/ops/blob/master/lepton/ldd_darwin.go#L75

cmd/cmd_env.go
```diff
@@ -0,0 +1,200 @@
...
+			goto cleanup
+		}
+		_, err = p.BuildImage(ctx)
+		if err != nil {
+			log.Errorf("Failed to build image: %v", err)
+		} else {
+			fmt.Printf("On-prem image '%s' created...\n", c.RunConfig.Imagename)
+		}
+	}
+cleanup:
```
eyberg, line 146: i'd convert this goto into a defer() that can be put at the top of the function

eyberg:
fyi, I made a local package out of this && ran it locally on mac and seems to work fine - so it's probably the difference between 'ops run' and the way 'pkg load' works

lepton/ldd_linux.go
```diff
@@ -69,6 +82,9 @@ func getSharedLibs(targetRoot string, path string) ([]string, error) {
 	if err == nil && IsDynamicLinked(elfFile) {
 		env := os.Environ()
 		env = append(env, "LD_TRACE_LOADED_OBJECTS=1")
+		if targetRoot != "" {
+			env = append(env, "LD_LIBRARY_PATH="+targetRoot)
+		}
```
eyberg, line 87: actually this might be what the problem is - this won't work on mac - i realize this file is for linux but will need something corresponding for ldd_darwin.go

francescolavra: I modified findLib() in ldd_darwin.go to take into account the target root, and now it works on mac too.

cmd/cmd_env.go
francescolavra: defer doesn't work when os.Exit() is called, so if we want ops to exit with a non-zero status code when an error is encountered (e.g. when the user specifies non-existent files in the config) we can't rely on defer.

eyberg: ah ok, nvr mind then

eyberg (approved):

PR #1136, opened 2021-08-03
initial support for instance groups on gcp

eyberg:


gcp/gcp_instance.go
```diff
@@ -20,6 +20,12 @@ func (p *GCloud) CreateInstance(ctx *lepton.Context) error {
 		c.CloudConfig.Flavor = "g1-small"
 	}
 
+	// short-circuit to instancegroup
+	if c.RunConfig.InstanceGroup != "" {
+		p.addToInstanceGroup(ctx, c.RunConfig.InstanceGroup)
+		return nil
```
francescolavra, line 26: It seem this would work, but any firewall rules that may be necessary to open TCP and UDP ports specified in the config wouldn't be created, and if the config specifies a domain name the relevant DNS records wouldn't be created either.

eyberg: domains would not be attached to instances in the instance group - it acts as an auto-scaling load balancer

eyberg: added ports for instance template

rrs45:
Wondering if OPS would ever replace Terraform or Ansible for VM
provisioning? If not then wouldn't it make sense to restrict Ops for image
creation only? Adding provisioning support in this manner will be never
ending process in my opinion which may not be worth in long term. Just
sharing my thoughts. Thanks

eyberg:
@rrs45 this is an optional feature and a direct customer request, also there's no reason why ops couldn't replace tf or ansible in many situations

francescolavra:
LGTM

PR #1146, opened 2021-08-07
Remove target root prefix path from library path

sanderssj:
When using the TargetRoot option where libraries must be added to the image,
the entire host path was used in the image. This change removes the TargetRoot
path from the host library path to create the correct path for the image.

### DIFF
--- a/fs/manifest.go
+++ b/fs/manifest.go
@@ -370,7 +370,7 @@ func (m *Manifest) AddFileTo(dir map[string]interface{}, filepath string, hostpa
 
 // AddLibrary to add a dependent library
 func (m *Manifest) AddLibrary(path string) {
-	parts := strings.FieldsFunc(path, func(c rune) bool { return c == '/' })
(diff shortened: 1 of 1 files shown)


fs/manifest.go
```diff
@@ -369,8 +369,8 @@ func (m *Manifest) AddFileTo(dir map[string]interface{}, filepath string, hostpa
 }
 
 // AddLibrary to add a dependent library
-func (m *Manifest) AddLibrary(path string) {
-	parts := strings.FieldsFunc(path, func(c rune) bool { return c == '/' })
+func (m *Manifest) AddLibrary(targetroot string, path string) {
+	parts := strings.FieldsFunc(strings.TrimPrefix(path, targetroot), func(c rune) bool { return c == '/' })
```
francescolavra, line 373: The Manifest type has a targetRoot field, so I would use that (as in `strings.TrimPrefix(path, m.targetRoot)`) instead of adding a parameter to AddLibrary().

sanderssj: Thanks for pointing that out!

francescolavra (approved):

PR #1157, opened 2021-09-09
external gcp ipv6 support

eyberg:


gcp/gcp_instance.go
```diff
@@ -115,7 +130,22 @@ func (p *GCloud) CreateInstance(ctx *lepton.Context) error {
 	}
 
 	if len(ctx.Config().RunConfig.UDPPorts) != 0 {
-		rule := p.buildFirewallRule("udp", ctx.Config().RunConfig.UDPPorts, instanceName)
+
+		if ctx.Config().CloudConfig.EnableIPv6 {
+			rule := p.buildFirewallRule("udp", ctx.Config().RunConfig.Ports, instanceName, true)
```
francescolavra, line 135: Should be `RunConfig.UDPPorts`

gcp/gcp_instance.go
```diff
@@ -235,15 +265,15 @@ func (p *GCloud) DeleteInstance(ctx *lepton.Context, instancename string) error
 	cloudConfig := ctx.Config().CloudConfig
 	runConfig := ctx.Config().RunConfig
 	if len(runConfig.Ports) != 0 {
-		rule := p.buildFirewallRule("tcp", runConfig.Ports, instancename)
+		rule := p.buildFirewallRule("tcp", runConfig.Ports, instancename, false)
 		_, err := p.Service.Firewalls.Delete(cloudConfig.ProjectID, rule.Name).Context(context).Do()
```
francescolavra, line 269: It would be better if IPv6 firewall rules are also deleted (for both TCP and UDP), otherwise if one deletes an instance and then recreates an instance with the same name as the previously deleted instance, they would get a "Failed to add Firewall rule" error.

eyberg: interesting - this actually relies on a config which probably isn't desirable since we know when these fw rules are created by ops; i added it here since this is the way it is but it should just auto-delete if created by ops, ill create a tkt

eyberg: interesting - this actually relies on a config which probably isn't desirable since we know when these fw rules are created by ops; i added it here since this is the way it is but it should just auto-delete if created by ops, ill create a tkt

PR #1159, opened 2021-09-21
use subnet for fw rules if passed in for gcp

eyberg:


gcp/gcp_network.go
```diff
@@ -208,16 +208,21 @@ func (p *GCloud) buildFirewallRule(protocol string, ports []string, tag string,
...
+	f := &compute.Firewall{
 		Name:         name,
 		Description:  fmt.Sprintf("Allow traffic to %v ports %s", arrayToString(ports, "[]"), tag),
 		Allowed:      allowed,
 		TargetTags:   []string{tag},
 		SourceRanges: []string{src},
 	}
+
+	if network != "" {
+		f.Network = "global/networks/" + network
```
francescolavra, line 222: According to https://pkg.go.dev/google.golang.org/api/compute/v1#Firewall:
> If you choose to specify this field, you can specify the network as a full or partial URL. For example, the following are all valid URLs:
> - https://www.googleapis.com/compute/v1/projects/myproject/global/networks/my-network
> - projects/myproject/global/networks/my-network
> - global/networks/default

So it appears that when specifying a non-default network you have to use either the full URL or the `projects/myproject/` prefix. In any case, apparently we need to retrieve the ProjectID from the CloudConfig.

francescolavra (approved):

PR #1162, opened 2021-09-24
fixing path prefix for Files config

eyberg:


lepton/image.go
```diff
@@ -239,7 +239,11 @@ func setManifestFromConfig(m *fs.Manifest, c *types.Config) error {
 	m.AddKlibs(c.RunConfig.Klibs)
 
 	for _, f := range c.Files {
-		hostPath := path.Join(c.LocalFilesParentDirectory, f)
+		hostPath := f
+		if string(f[0]) != "/" {
+			hostPath = path.Join(c.LocalFilesParentDirectory, f)
+		}
+
```
francescolavra, line 246: If a target root path is specified in the configuration, it should be taken into account here. In the past I've had problems when trying to insert into an image files from a specific target root (e.g. when the target root is the directory where files from a crossbuild environment have been placed), and I used the following patch to make it work:
```
        m.AddKlibs(c.RunConfig.Klibs)
 
        for _, f := range c.Files {
-               hostPath := path.Join(c.LocalFilesParentDirectory, f)
+               var hostPath string
+               if filepath.IsAbs(f) {
+                       hostPath = filepath.Join(c.TargetRoot, f)
+               } else {
+                       hostPath = path.Join(c.LocalFilesParentDirectory, f)
+               }
                filePath := f
                err := m.AddFile(filePath, hostPath)
                if err != nil {
```

francescolavra (approved):

PR #1167, opened 2021-10-04
seems DO doesnt like our raw image; converting to qcow2 seems to fix …

eyberg:
…the issue

### DIFF
--- a/digitalocean/digital_ocean_image.go
+++ b/digitalocean/digital_ocean_image.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


digitalocean/digital_ocean_image.go
```diff
@@ -46,11 +54,13 @@ func (do *DigitalOcean) CreateImage(ctx *lepton.Context, imagePath string) error
...
 
-	_, _, err = do.Client.Images.Create(context.TODO(), createImageRequest)
+	i, _, err := do.Client.Images.Create(context.TODO(), createImageRequest)
 	if err != nil {
 		return err
 	}
 
+	log.Infof("%+v\n", i)
+
 	err = do.Storage.DeleteFromBucket(c, filepath.Base(imagePath))
```
sanderssj, line 64: I think this should be newPath instead of imagePath

sanderssj (commented):
LGTM other than the path for bucket file deletion

sanderssj:
Actually, when doing some testing with deleting the changed file name, the image often is not created. Not sure if that means the image isn't always created by the time the file gets deleted.

sanderssj (approved):

sanderssj:
There is a problem where the image may not be created when the file is deleted too quickly. The fix for this is to loop querying the image id until its status becomes "available", e.g.:

```
	for i.Status != "available" {
		time.Sleep(100 * time.Millisecond)
		i, _, err = do.Client.Images.GetByID(context.TODO(), i.ID)
		if err != nil {
			return err
		}
	}
```

PR #1169, opened 2021-10-05
add bash auto-complete compat for zsh

eyberg:


sanderssj:
I played around with this a little bit and it didn't work for me unless I ran bashcompinit. I'm not really sure what the best approach here is since it seems the zsh user is intended to set up the completions their own way, and scripts are not encouraged to run compinit on their own as running it again will undo any special invocations the user may have used. I experimented with autoload +X, but that doesn't seem to work as expected when it's sourced from a different script.

I'll throw out an alternative approach: moving the zsh check into the bash autocomplete script and make the running of compinit and bashcompinit dependent upon the functions they create existing. So, in bash_completion.sh, something like:
```
if [ "$(basename $SHELL)" = "zsh" ]; then
        if ! command -v compdef > /dev/null; then
                autoload -Uz compinit && compinit
        fi
        if ! command -v complete > /dev/null; then
                autoload -Uz bashcompinit && bashcompinit
        fi
fi
```

If the user has neither set up in their .zshrc or site-wide location, then it will work, and if they have it already set up then it should not undo any special invocation they used for it. Alternatively/additionally, we could just echo a message asking the user to set their .zshrc with these commands? What do you think?

eyberg:
i'm not having this problem personally but I also have a ton of other plugins that probably already handle this, i kind of agree that this isn't necessarily something ops should be doing but i'm not opposed to including it either since we do make changes in other places like modifying path

@optimizasean - any preference?

optimizasean:
Ran some tests on my local machine and a few other machines, everything seems to work out of the box now!  Thanks!

PR #1211, opened 2021-12-21
Use aarch64 url when updating ops on arm

sanderssj:


eyberg (approved):
nice - i just ran into this myself

PR #1216, opened 2021-12-24
un-stub kernel in local ops run/build for arm64

eyberg:


francescolavra (approved):
This doesn't support running a nightly version, or a local kernel binary specified in config.json, but in order to support that we would have to refactor the code so that the entire config (and not just the RunConfig part) is available to the setConfig() function, because the path to the nightly (or an arbitrary local) kernel binary would be set there.

eyberg:
yeh, this is just a tmp. thing cause we're still passing in '-kernel' in; eventually if/when we want to support something like graviton instances we need to get rid of that

Issue #1116, opened 2021-07-09
pkg from-docker could use an option to cp the entire fs vs just the program in question and it's linked libs


eyberg:
i kinda like the way it works now but typically many programs will have other files that are necessary in order to function like config files and such

maybe just have a switch to cp the fs into sysroot

quantumsheep:
I though of a `--copy`/`-c` option, there are cases where I wanted to to that as well.

girishramnani:
We can introduce the `-c` `--copy` option like mentioned by @quantumsheep and have a extra optional argument to configure the max depth of folders to copy which can be `2` level as defaults?

eyberg:
i don't see a need for specifying the tree depth to copy; I think the default case is that people will want everything; it is just you can get by with very little by copying only the program; i'd maybe just add the copy for now and then we can play with it and change it later if need

girishramnani:
so basically `--copy` will take a directory as an input? 
example - `ops pkg from-docker node:16.3.0 -f node -c ./` or should it also work when someone provides just a file like
`ops pkg from-docker node:16.3.0 -f node -c package.json` @eyberg ?

eyberg:
no, no input.

right now using the node example it simply grabs the node binary and the libraries linked to it and that's it; this works for a lot of programs, however, there is typically quite a lot of other stuff in the docker image though (could be config files, locales, etc.) so the proposed copy should just export the entire docker image into the unikernel output

girishramnani:
oh ok makes sense
So I should export the whole fs of the docker image to the unikernel? or just the working directory of the docker image @eyberg ?

girishramnani:
seems like copying the whole fs is the way to go as per the issue description

girishramnani:
Working on this task

eyberg:
whole fs

PR #1221, opened 2022-01-03
implement copy flag for from-docker command

girishramnani:
Closes https://github.com/nanovms/ops/issues/1116

### DIFF
--- a/cmd/extract_from_docker.go
+++ b/cmd/extract_from_docker.go
@@ -316,7 +316,6 @@ func copyWholeContainer(cli *dockerClient.Client, containerID string, hostBaseDi
 		}
 
 		target := filepath.Join(hostBaseDir, header.Name)
-		fmt.Println(target)
 
 		// check the file type
 		switch header.Typeflag {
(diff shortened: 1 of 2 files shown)


cmd/extract_from_docker.go
```diff
@@ -272,6 +283,68 @@ func copyFromContainer(cli *dockerClient.Client, containerID string, containerPa
...
+		case err != nil:
+			return err
+
+		// if the header is nil, just skip it (not sure how this happens)
+		case header == nil:
+			continue
+		}
+
+		target := filepath.Join(hostBaseDir, header.Name)
+		fmt.Println(target)
```
eyberg, line 319: i'd only show this if debug/log verbosity is turned on

eyberg (approved):

cmd/extract_from_docker.go
girishramnani: Gotcha, forgot to remove that

girishramnani: removed it

PR #1223, opened 2022-01-04
initial freebsd support

eyberg:
this still relies on qemu for now but will run hello worlds


sanderssj:
Does the ldd_freebsd.go actually work? Wouldn't the ldd_darwin.go code be more appropriate since you would need a linux target root to pull from?

eyberg:
nah, it's just copied right now, i'll just stub them out instead

sanderssj (approved):
Sounds good

PR #1227, opened 2022-01-10
fix issues/1087

haitham911:


lepton/image.go
```diff
@@ -254,6 +255,12 @@ func setManifestFromConfig(m *fs.Manifest, c *types.Config) error {
 	}
 
 	for k, v := range c.MapDirs {
+		for _, x := range c.Args {
+			if x == filepath.Base(v) {
+				errstr := fmt.Sprintf("can't have directory with same name as binary %v", x)
+				return errors.New(errstr)
```
eyberg, line 261: instead of the panic - i'd just make this error out with this msg and exit

PR #1239, opened 2022-01-23
create IAM role when create instance

### LINKED ISSUE #1231: allow user to add iam role via config
https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.AssociateIamInstanceProfile

olegpono:
Fixed #1231

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,7 +1,0 @@
-{
-  "CloudConfig" :{
-    "ProjectID" :"projectid",
-    "BucketName":"thebucketname"
-  },
-  "RunConfig" : {"Memory": "2G"}
-}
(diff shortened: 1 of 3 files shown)


config.json
```diff
@@ -1,7 +1,29 @@
 {
```
eyberg, line 1: i'm not sure why this config.json is being checked in - we should just remove this file entirely

types/config.go
```diff
@@ -162,6 +162,39 @@ type ProviderConfig struct {
...
 	// gcp: https://cloud.google.com/compute/docs/regions-zones#available
 	Zone string `cloud:"zone"`
+
+	// AwsIAM is used to define IAM Role AWS
+	AwsIAM *AwsIAM
+}
+
+// AwsIAM define IAM Role AWS
+type AwsIAM struct {
+	// RoleName
```
eyberg, line 172: the structs need to have comments since they are being exported but the struct fields themselves don't need comments unless there is something to mention; repeating the name doesn't really add anything - maybe just remove these?

eyberg (approved):

Issue #1230, opened 2022-01-18
if json fails to load lint to show position it fails at


eyberg:
```Error: failed converting configuration file: invalid character ',' after object key```

**Comment by eyberg** [comment]
the idea here is to lint the json and inform the user on what line/char they might have invalid json

for example this ->

```
{
	"CloudConfig": {
		"ProjectID": "projectid",
		"BucketName": "thebucketname"
	},
	total garbage
         "RunConfig": {
		"Memory": "2G"
	}
}
```

might produce an error msg like so:

```
Error: Parse error on line 5:
...thebucketname"	},	total garbage "RunCo
---------------------^
Expecting 'STRING', got 'undefined'
```

PR #1241, opened 2022-01-23
1230/Show position in code if json config fails to load

patche-v:
Fixed [1230](https://github.com/nanovms/ops/issues/1230)

### DIFF
--- a/cmd/flags_config.go
+++ b/cmd/flags_config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path"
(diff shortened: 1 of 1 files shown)


eyberg (changes requested):
sorry, that ticket isn't super clear - we don't necessarily care about the position in ops if the json is invalid - we'd like to see the line/char that the json fails at, i updated the tkt w/an example

eyberg (approved):

eyberg:
thanks!

PR #1260, opened 2022-01-31
add auto scaling group when creating an instance

### LINKED ISSUE #1137: add aws auto scaling groups
aws equivalent of https://github.com/nanovms/ops/issues/1135

olegpono:
Fixed #1137

### DIFF
--- a/aws/aws_instance.go
+++ b/aws/aws_instance.go
@@ -198,6 +198,7 @@ func (p *AWS) CreateInstance(ctx *lepton.Context) error {
 
 	svc := p.ec2
 	cloudConfig := ctx.Config().CloudConfig
+	runConfig := ctx.Config().RunConfig
 
 	// create security group - could take a potential 'RemotePort' from
 	// config.json in future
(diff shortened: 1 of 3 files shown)


types/config.go
```diff
@@ -165,6 +165,9 @@ type ProviderConfig struct {
 
 	// AwsIPN is aws instance profile name
 	AwsIPN string
+
+	// AwsASG is aws auto scaling group
+	AwsASG string
```
eyberg, line 170: dupe?

types/config.go
```diff
@@ -206,6 +209,12 @@ type RunConfig struct {
 	// InstanceGroup
 	InstanceGroup string
 
+	// AwsIPN is Aws instance profile name
```
eyberg, line 212: we'd like these to be cloud agnostic as much as possible so should drop the 'aws' prefix

types/config.go
```diff
@@ -206,6 +209,12 @@ type RunConfig struct {
...
+	AwsIPN string
+
+	// AwsASG is Aws auto scaling group
+	AwsASG string
```
eyberg, line 216: I think InstanceGroup is the equivalent being used on GCP so we should use that instead

aws/aws_instance.go
```diff
@@ -284,9 +285,47 @@ func (p *AWS) CreateInstance(ctx *lepton.Context) error {
...
+			instanceNIS.Ipv6Addresses = []*ec2.InstanceIpv6Address{addie}
 		} else {
-			instanceInput.SetIpv6AddressCount(1)
+			instanceNIS.SetIpv6AddressCount(1)
+		}
+	}
+
+	instanceInput.NetworkInterfaces = []*ec2.InstanceNetworkInterfaceSpecification{instanceNIS}
+
+	if cloudConfig.AwsIPN != "" || runConfig.AwsIPN != "" {
```
eyberg, line 296: is this actually required for scaling groups? I don't think so?

aws/aws_instance.go
```diff
@@ -253,29 +254,29 @@ func (p *AWS) CreateInstance(ctx *lepton.Context) error {
 	tags, tagInstanceName := buildAwsTags(cloudConfig.Tags, ctx.Config().RunConfig.InstanceName)
 	tags = append(tags, &ec2.Tag{Key: aws.String("image"), Value: &imgName})
 
+	instanceNIS := &ec2.InstanceNetworkInterfaceSpecification{
+		AssociatePublicIpAddress: aws.Bool(true),
```
eyberg, line 258: this shouldn't be hard-coded

eyberg (changes requested):
i'd make all those config names cloud-agnostic - for instance we already have' instance group' for gcp

aws/aws_instance.go
olegpono: Aws IntanceProfileName is optional

cmd/cmd_instance.go
```diff
@@ -48,6 +48,8 @@ func instanceCreateCommand() *cobra.Command {
 	PersistCreateInstanceFlags(cmdInstanceCreate.PersistentFlags())
 	cmdInstanceCreate.PersistentFlags().StringP("instance-name", "i", "", "instance name (overrides default instance name format)")
 	cmdInstanceCreate.PersistentFlags().StringP("instance-group", "", "", "instance group")
+	cmdInstanceCreate.PersistentFlags().StringP("aws-ipn", "", "", "AWS instance profile name")
+	cmdInstanceCreate.PersistentFlags().StringP("aws-asg", "", "", "AWS auto scaling group")
```
eyberg, line 52: this should be the same as instance group right?

cmd/cmd_instance.go
```diff
@@ -48,6 +48,8 @@ func instanceCreateCommand() *cobra.Command {
 	PersistCreateInstanceFlags(cmdInstanceCreate.PersistentFlags())
 	cmdInstanceCreate.PersistentFlags().StringP("instance-name", "i", "", "instance name (overrides default instance name format)")
 	cmdInstanceCreate.PersistentFlags().StringP("instance-group", "", "", "instance group")
+	cmdInstanceCreate.PersistentFlags().StringP("aws-ipn", "", "", "AWS instance profile name")
```
eyberg, line 51: if we have the equivalent of the GCP version of this we should use this and if not just make it agnostic so it's not aws specific

eyberg (approved):

PR #1263, opened 2022-02-02
fix help text for ops volume attach

girishramnani:
the current help for `ops volume attach` seems to be incorrect as it shows the first arg to be `image_name` but really expects `instance_name` so changing that.
Also changed the instance argument names to reflect this change.

### DIFF
--- a/lepton/provider.go
+++ b/lepton/provider.go
@@ -61,11 +61,11 @@ type Storage interface {
 
 // VolumeService is an interface for volume related operations
 type VolumeService interface {
-	CreateVolume(ctx *Context, name, data, provider string) (NanosVolume, error)
+	CreateVolume(ctx *Context, volumeName, data, provider string) (NanosVolume, error)
 	GetAllVolumes(ctx *Context) (*[]NanosVolume, error)
-	DeleteVolume(ctx *Context, name string) error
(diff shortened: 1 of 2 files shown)


lepton/provider.go
```diff
@@ -61,11 +61,11 @@ type Storage interface {
...
 type VolumeService interface {
-	CreateVolume(ctx *Context, name, data, provider string) (NanosVolume, error)
+	CreateVolume(ctx *Context, volumeName, data, provider string) (NanosVolume, error)
 	GetAllVolumes(ctx *Context) (*[]NanosVolume, error)
-	DeleteVolume(ctx *Context, name string) error
-	AttachVolume(ctx *Context, image, name string) error
-	DetachVolume(ctx *Context, image, name string) error
+	DeleteVolume(ctx *Context, volumeName string) error
+	AttachVolume(ctx *Context, instanceName, volumeName string) error
+	DetachVolume(ctx *Context, instainceName, volumeName string) error
```
eyberg, line 68: sp

girishramnani: good catch

girishramnani: fixed

eyberg (approved):

girishramnani:
This can be merged

PR #1266, opened 2022-02-04
aws graviton instances set Uefi flag when BuildImage

### LINKED ISSUE #1249: for graviton instances set Uefi flag
```
    "Uefi": true,
```

olegpono:
Fixed #1249

### DIFF
--- a/aws/aws_image.go
+++ b/aws/aws_image.go
@@ -30,6 +30,11 @@ import (
 // BuildImage to be upload on AWS
 func (p *AWS) BuildImage(ctx *lepton.Context) (string, error) {
 	c := ctx.Config()
+
+	if isGravitonProcessor(c.CloudConfig.Flavor) {
+		c.Uefi = true
+	}
+
 	err := lepton.BuildImage(*c)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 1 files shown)


aws/aws_image.go
```diff
@@ -322,6 +330,15 @@ func getArchitecture(flavor string) string {
 	return "x86_64"
 }
 
+func isGravitonProcessor(flavor string) bool {
+	if flavor == "" {
+		return false
+	}
+
+	_, isExist := GravitonProcessors[strings.ToLower(strings.Split(flavor, ".")[0])]
```
eyberg, line 338: is there a flavor with a period in it? it doesn't look like it? do we need this split here?

eyberg (changes requested):

aws/aws_image.go
olegpono: yes, the flavor has a period, we need split here for GravitonProcessors mapping like this: https://github.com/nanovms/ops/blob/a69fa8582afb15b16156c1fe1e629ecd4708dc71/aws/aws_image.go#L308
example: r6g.medium, a1.xlarge

eyberg (approved):

PR #1270, opened 2022-02-11
support for new pkghub manifest format which uses namespaced packages

girishramnani:
- support for new pkghub manifest format which uses namespaced packages


eyberg (approved):
need to fix the govet/golint stuff

Issue #1273, opened 2022-02-15
c6a instance type belongs in nitro instance types map


eyberg:
"Flavor": "c6a.xlarge", needs ena adapter

https://github.com/nanovms/ops/blob/69a47b51ef729ab54526c73806edbddd2b61f36c/aws/aws_image.go#L347

PR #1274, opened 2022-02-16
add support for the new c6a instance type

zeroecco:
add basic support for c6a instance types so ENA works


zeroecco:
for tkt: https://github.com/nanovms/ops/issues/1273

eyberg (approved):

PR #1286, opened 2022-03-07
fix local package listing

girish-nanovms:
This is how it would look like post this PR change
```
+---------------+-------------+----------------+----------+---------+-------------+
|   NAMESPACE   | PACKAGENAME |    VERSION     | LANGUAGE | RUNTIME | DESCRIPTION |
+---------------+-------------+----------------+----------+---------+-------------+
| g-nanos | b2b         | b2b            |          |         |             |
+---------------+-------------+----------------+----------+---------+-------------+
| g-nanos | node        | node-16.3.0    | c        | c       | hi          |
+---------------+-------------+----------------+----------+---------+-------------+
```

You can see that once i added the runtime, language and description in the package.manifest they are picked up for `node-16.3.0`

### DIFF
--- a/lepton/package.go
+++ b/lepton/package.go
@@ -250,9 +250,10 @@ func GetLocalPackageList() ([]Package, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	username := GetLocalUsername()
 	for _, pkg := range localPackages {
 		pkgName := pkg.Name()
+		pkgParts := strings.Split(pkg.Name(), "-")
 
 		// ignore packages compressed
 		if !strings.Contains(pkgName, "tar.gz") {
@@ -268,7 +269,8 @@ func GetLocalPackageList() ([]Package, error) {
 				os.Exit(1)
 				return nil, err
 			}
-
+			pkg.Namespace = username
+			pkg.Name = pkgParts[0]
 			packages = append(packages, pkg)
 		}
 	}
--- a/lepton/package.go
+++ b/lepton/package.go
@@ -3,6 +3,7 @@ package lepton
 import (
 	"archive/tar"
 	"crypto/sha256"
+	"regexp"
 
 	"compress/gzip"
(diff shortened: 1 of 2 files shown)


lepton/package.go
```diff
@@ -250,9 +250,10 @@ func GetLocalPackageList() ([]Package, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	username := GetLocalUsername()
 	for _, pkg := range localPackages {
 		pkgName := pkg.Name()
+		pkgParts := strings.Split(pkg.Name(), "-")
```
girish-nanovms, line 256: as `from-docker` uses a `-` to join the image name and version, I felt that `-` would be appropriate to split the version

eyberg: this is better but this is only for the case where you import 'from-docker' - any package not imported via that (eg: you create yourself or create from a directory),

so all 3 of these are 'valid' and produce:

python_3.8.6 -> 3.8.6
node-16.3.0 -> 16.3.0
nginx -> ''

girish-nanovms: ok so how would we handle `java_1.8.0_190` as we have such posibility. maybe be consider everything after the first `_` as version?

girish-nanovms: ok maybe I can regex it

eyberg: yes for now just use the first underscore - looking towards the future this probably doesn't belong in the actual filename itself

girish-nanovms: done

eyberg: so this is still populating packages without versions the name of the package which doesn't make sense - i'd just leave it empty string if there is no version;

also i have a package named 'node-16.3.0' and that is populating the version as 'node-16.3.0'

eyberg: in general a version really shouldn't have anything but numerics && periods

Issue #1109, opened 2021-07-01
gcp instance list shows blank image column


eyberg:
```ops instance list```

PR #1293, opened 2022-03-10
add image label to gcp instance on creation

girish-nanovms:
Closes https://github.com/nanovms/ops/issues/1109

Note - this will not help any existing instance and will only work for instances created after this PR is merged.


girish-nanovms:
Verified that this works correctly

eyberg (approved):

Issue #874, opened 2021-01-22
add optional net console flag


eyberg:
https://github.com/nanovms/nanos/pull/1381

girish-nanovms:
how should the flag look like?

girish-nanovms:
`--netconsole_port`, `--netconsole_ip` and `--consoles` optional args?

girish-nanovms:
Also what all commands should have these flags?

PR #1297, opened 2022-03-13
add optional net console flag to commands

girish-nanovms:
closes https://github.com/nanovms/ops/issues/874


francescolavra (commented):
In my opinion, these flags should be incorporated in the existing BuildImageCommandFlags instead of creating a new NetConsoleFlags type, since they affect how an image is built and apply to the same commands to which the BuildImageCommandFlags apply. Besides, the `--consoles` flag is not specific to the netconsole feature.

girish-nanovms:
opening a new PR

PR #1302, opened 2022-03-23
fixing pkg path for image create pkgs

eyberg:


zeroecco:
This worked for my issue

cmd/cmd_image.go
```diff
@@ -72,11 +72,16 @@ func imageCreateCommandHandler(cmd *cobra.Command, args []string) {
 	providerFlags := NewProviderCommandFlags(flags)
 	pkgFlags := NewPkgCommandFlags(flags)
 
+	if pkgFlags.Package != "" {
+		pkgFlags.SluggedPackage = strings.ReplaceAll(pkgFlags.Package, ":", "_")
+	}
```
francescolavra, line 77: This would need to be added to deployCommandHandler() as well, in cmd/cmd_deploy.go.
Alternatively, to avoid code duplication we could have this done in a single place, e.g. in MergeToConfig() in cmd/flags_pkg.go.

PR #1305, opened 2022-03-24
add netconsole flags and config passthrough

girish-nanovms:
closes https://github.com/nanovms/ops/issues/874

### DIFF
--- a/lepton/image.go
+++ b/lepton/image.go
@@ -39,22 +39,6 @@ func BuildImage(c types.Config) error {
 	return nil
 }
 
-// rebuildImage rebuilds a unikernel image for user
-// supplied ELF binary after volume attach/detach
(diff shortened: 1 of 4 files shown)


lepton/image.go
```diff
@@ -39,22 +39,6 @@ func BuildImage(c types.Config) error {
 	return nil
 }
 
-// rebuildImage rebuilds a unikernel image for user
-// supplied ELF binary after volume attach/detach
-func rebuildImage(c types.Config) error {
```
eyberg, line 44: looks like this and the refreshRate were removed because they aren't referenced which is fine but for future ref prob. best to include this in a sep. pr to eliminate any confusion

eyberg (approved):

francescolavra (commented):
Looks good to me

PR #1313, opened 2022-04-05
fix 'no local package of name found' error when pushing local packages

girish-nanovms:
currently when you create a package locally using `from-docker` or manually and try to upload it to `repo.ops.city` then ops throws error `no local package with the name <name> found` even if the package is present locally.

This PR fixes the above bug.


eyberg (approved):

francescolavra (commented):
Looks good to me, once the golint error in CI is fixed

PR #1323, opened 2022-05-08
Add feature to upload public and private packages to organization namespaces

girish-nanovms:
This change adds the support for uploading to org namespace. The way it works is that now you can provide an optional namespace prefix to your local folder
```
ops pkg push test/golang-1.8.1
```
Here ops will use the `test` as namespace and `golang-1.8.1` would be folder name it will search in `local_package`.


eyberg (approved):

girish-nanovms:
This PR also works with private non namespaced packages

PR #1324, opened 2022-05-10
fuzzy match version when pushing package and some fixes

girish-nanovms:
Fixes the version mismatch when `v` is set as prefix.
`node_v16.5.0` folder but manifest says `16.5.0`. So now we fuzzy match the version.

Also only read local manifest if it exists

### DIFF
--- a/cmd/cmd_pkg.go
+++ b/cmd/cmd_pkg.go
@@ -366,13 +366,14 @@ func cmdPkgPush(cmd *cobra.Command, args []string) {
 			log.Fatal(err)
 		}
 	}
-
 	private, _ := flags.GetBool("private")
 	ns, name, version := api.GetNSPkgnameAndVersion(pkgIdentifier)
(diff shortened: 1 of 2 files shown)


cmd/cmd_pkg.go
```diff
@@ -366,20 +366,21 @@ func cmdPkgPush(cmd *cobra.Command, args []string) {
...
 		}
 	}
-
 	private, _ := flags.GetBool("private")
 	ns, name, version := api.GetNSPkgnameAndVersion(pkgIdentifier)
 	pkgList, err := api.GetLocalPackageList()
 	if err != nil {
 		log.Fatal(err)
 	}
+	fmt.Println(ns, name, version)
```
eyberg, line 375: can rm now

eyberg (approved):

cmd/cmd_pkg.go
girish-nanovms: done

PR #1344, opened 2022-06-14
I see no reason to suffix these images if we arent even using the cor…

eyberg:
…rect suffix to mention type


francescolavra (commented):
With this change, there would be an inconsistency between 'ops image list` (which would list all images, with or without the .img extension) and 'ops instance create' (which errors out with "image not found" when trying to create an instance from an image without the extension). I would suggest the additional changes below, which would allow creating instances from any of the images listed by 'ops image list':
```
diff --git a/onprem/onprem_instance.go b/onprem/onprem_instance.go
index d5e48a7..0de3111 100644
--- a/onprem/onprem_instance.go
+++ b/onprem/onprem_instance.go
@@ -22,14 +22,16 @@ import (
 func (p *OnPrem) CreateInstance(ctx *lepton.Context) error {
        c := ctx.Config()
 
-       imageName := c.CloudConfig.ImageName
-
-       if matched, _ := regexp.Match(`.img$`, []byte(imageName)); !matched {
-               c.CloudConfig.ImageName = imageName + ".img"
-       }
+       opshome := lepton.GetOpsHome()
+       imgpath := path.Join(opshome, "images", c.CloudConfig.ImageName)
 
-       if _, err := os.Stat(path.Join(lepton.GetOpsHome(), "images", c.CloudConfig.ImageName)); os.IsNotExist(err) {
-               return fmt.Errorf("image \"%s\" not found", imageName)
+       if _, err := os.Stat(imgpath); os.IsNotExist(err) {
+               if matched, _ := regexp.Match(`.img$`, []byte(c.CloudConfig.ImageName)); !matched {
+                       imgpath = imgpath + ".img"
+               }
+               if _, err := os.Stat(imgpath); os.IsNotExist(err) {
+                       return fmt.Errorf("image \"%s\" not found", c.CloudConfig.ImageName)
+               }
        }
 
        if c.Mounts != nil {
@@ -53,9 +55,6 @@ func (p *OnPrem) CreateInstance(ctx *lepton.Context) error {
 
        fmt.Printf("booting %s ...\n", c.RunConfig.InstanceName)
 
-       opshome := lepton.GetOpsHome()
-       imgpath := path.Join(opshome, "images", c.CloudConfig.ImageName)
-
        c.RunConfig.Imagename = imgpath
        c.RunConfig.Background = true
 
```

cmd/cmd_image.go
```diff
@@ -661,8 +661,7 @@ func getLocalImageReader(flags *pflag.FlagSet, args []string) *fs.Reader {
 	imageName := args[0]
 	imagePath := path.Join(api.LocalImageDir, imageName)
 	if _, err := os.Stat(imagePath); err != nil {
-		if os.IsNotExist(err) && !strings.HasSuffix(imagePath, ".img") {
-			imagePath += ".img"
+		if os.IsNotExist(err) {
 			_, err = os.Stat(imagePath)
```
francescolavra, line 665: This `os.Stat(imagePath)` call was there to check for the existence of an image file after adding the .img suffix; if we don't use the suffix anymore, we could avoid this call altogether, since it's redundant with the previous one at line 663.

onprem/onprem_instance.go
```diff
@@ -24,9 +23,7 @@ func (p *OnPrem) CreateInstance(ctx *lepton.Context) error {
 
 	imageName := c.CloudConfig.ImageName
 
-	if matched, _ := regexp.Match(`.img$`, []byte(imageName)); !matched {
-		c.CloudConfig.ImageName = imageName + ".img"
-	}
+	c.CloudConfig.ImageName = imageName
```
francescolavra, line 26: This line can be removed altogether, since we have `imageName := c.CloudConfig.ImageName` above

PR #1345, opened 2022-06-20
Fixing grammar/punctuation in UNIKERNELS.md

LiberlandHacker:
# Changes should be pretty self-explanatory. 

1. Adding a "?" to the end of the H1 since it is ultimately used as a question.
2. Hyphenating "single-process" when it is used as a modifier, but not when it could be interpreted as the subject.
3. Capitalizing proper nouns like "Ruby" and "Unix". 
4. ...and other minor details.

### DIFF
--- a/UNIKERNELS.md
+++ b/UNIKERNELS.md
@@ -1,30 +1,30 @@
-# What is a Unikernel
+# What is a Unikernel?
 
-A unikernel is a single process system that is specifically designed to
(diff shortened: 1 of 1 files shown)


eyberg (approved):

eyberg:
thanks!

LiberlandHacker:
> thanks!

You're very welcome!

PR #1364, opened 2022-08-30
Add checks for data results from ProxMox API

asyslinux:
Added functions with checks {"data": null} or {"data":[]} results from ProxMox API.
This is need for checking permissions in proxmox, correctness environment variables etc.
It is required in order to stop the ops giving the result as if everything is fine, but in fact nothing works with proxmox.
Currently in error messages uses 'local' and 'local-lvm' hard-coded words, later when we have customizable name of storages, this words be replaced by variables.

### DIFF
--- a/proxmox/proxmox_checks.go
+++ b/proxmox/proxmox_checks.go
@@ -1,0 +1,140 @@
+package proxmox
+
+import (
+	"bytes"
+	"crypto/tls"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+type pData struct {
+	Data string `json:"data"`
+}
+
+type pDataArr struct {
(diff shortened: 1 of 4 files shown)


proxmox/proxmox_checks.go
```diff
@@ -0,0 +1,140 @@
...
+	var err error
+
+	edef := "check token permissions, environment variables correctness"
+	ecs := errors.New(edef)
+
+	switch rtype {
+	case "createimage":
+		ecs = errors.New("can not create disk image in 'local' storage, " + edef)
+	case "listimages":
+		ecs = errors.New("no have any disk images in 'local' storage or " + edef)
```
eyberg, line 94: sp: 'no disk images found in...'

eyberg (approved):

PR #1368, opened 2022-09-02
Added error checking coming from ProxMox API

asyslinux:
Added capture errors from ProxMox API.
Unfortunately, ProxMox does not always return an error, for example, in the case of a disk move, if there is no storage local-lvm, then the error is not returned. ProxMox doesn't have the best API.


eyberg (approved):

eyberg:
yeh I think that's why i had originally punted on that - thanks for adding!

PR #1369, opened 2022-09-02
Added forced checks for availability of storages in ProxMox before cr…

asyslinux:
Added forced checks for availability of storages in ProxMox before creating images and virtual machines.
Since ProxMox does not always return errors, for example, in the case of moving disks, I decided to make a forced check for the availability of storages, including their status.

### DIFF
--- a/proxmox/proxmox_checks.go
+++ b/proxmox/proxmox_checks.go
@@ -18,6 +18,16 @@ type pDataArr struct {
 	Data []string `json:"data"`
 }
 
+type sData struct {
+	Active  int    `json:"active"`
+	Enabled int    `json:"enabled"`
+	Storage string `json:"storage"`
+}
+
+type sDataArr struct {
+	Data []sData `json:"data"`
+}
+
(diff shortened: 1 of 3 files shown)


proxmox/proxmox_checks.go
```diff
@@ -63,6 +74,78 @@ func (p *ProxMox) CheckInit() error {
...
+// CheckStorage return error when not found configured storage or any storages via ProxMox API
+func (p *ProxMox) CheckStorage(storage string) error {
+
+	var err error
+
+	var sdp sDataArr
+
+	var b bytes.Buffer
+
+	edk := errors.New("no any storages is configured")
```
eyberg, line 86: if these are used only once maybe worth moving into their respective err handling blocks so they aren't instantiated each time?

eyberg (approved):

proxmox/proxmox_checks.go
asyslinux: I think I will still optimize and refactor the code about errors in proxmox handlers. I also think to make http client in one function, just substitute urls.

Issue #1356, opened 2022-08-28
Can not use ops with ProxMox 7.x


asyslinux:
Hello, I try to use ops with local ProxMox 7.2 and Ops 0.1.32, and it isn`t working:

1. I create a simple GO app
```
root@my:~/go/src/test# ls -la
total 6472
drwxr-xr-x 3 root root    4096 Aug 28 05:03 .
drwxr-xr-x 3 root root    4096 Aug 28 04:44 ..
-rw-r--r-- 1 root root     430 Aug 28 05:03 config.json
-rw-r--r-- 1 root root      21 Aug 28 04:46 go.mod
-rw-r--r-- 1 root root     354 Aug 28 04:46 main.go
drwxr-xr-x 2 root root    4096 Aug 28 05:00 static
-rwxr-xr-x 1 root root 6601074 Aug 28 04:48 test
```
2. Create a config.json:
```
{
    "Dirs": ["static"],
    "NameServers": ["172.21.1.1"],
    "RunConfig": {
        "Imagename": "test",
        "InstanceName": "1000",
        "Memory": "2G",
        "VolumeSizeInGb": 1,
        "TapName": "tap0",
        "Bridged": true,
        "BridgeName": "vmbr1",
        "IPAddress": "172.21.1.75",
        "NetMask": "255.255.255.0",
        "Gateway": "172.21.1.1",
        "Ports": ["80", "8008"],
        "Background": false,
        "Verbose": true,
        "ShowErrors": true,
        "ShowWarnings": true
    }
}
```
3. Try to create/build image:

```
root@my:~/go/src/test#ops image create test -t proxmox -c config.json
 100% |████████████████████████████████████████|  [1s:0s]
PWD:  /root/go/src/test
open /root/.ops/images/test.img: no such file or directory
```
```
root@my:~/go/src/test#ops build test -i test -t proxmox -c config.json
 100% |████████████████████████████████████████|  [1s:0s]
Bootable image file:
```
```
root@my:~/go/src/test# ops --show-debug image create test -t proxmox -c config.json
 100% |████████████████████████████████████████|  [1s:0s]
PWD:  /root/go/src/test
open /root/.ops/images/test.img: no such file or directory
```

Access rigths is full on token in ProxMox.
Image can not be build, nothing any errors from ops. 

```
root@my:~/go/src/test# ops image list -t proxmox
+----+------+------+
| ID | NAME | SIZE |
+----+------+------+
```

```
root@my:~/go/src/test# ops version
[...]

Issue #1009, opened 2021-04-22
Deployed example code doesn't accept requests from within a DigitalOcean VPC


willdavidc:
Hi, I've been investigating the possible use of ops in a production environment, on DigitalOcean (DO). I've been trying to get the minimal example from the README working properly.

I followed the instructions, created `hi.js`, built an image with `ops image create --package node_v14.2.0 -a hi.js`, uploaded to DO and spun up a droplet based on the image. This all worked great, and when I tried to access the public IP of the instance on port 8083, I got the expected `Hello World` response.

However, DO also assigns a private VPC IP to each host, in this case, `10.120.0.7`, and the image doesn't respond to requests there. This includes health checks from DO load balancers (which can only use the private IP address), and when I spun up a linux box within the VPC and `curl`ed the address:port from there.

<img width="503" alt="Screen Shot 2021-04-22 at 12 39 53 PM" src="https://user-images.githubusercontent.com/14321983/115752193-de366d80-a367-11eb-8502-39c96b752c82.png">

Some limitations of DO's system:
- No way to know in advance what the IP addresses of a new host will be (their "floating IPs" can't be assigned to droplets with custom images)
- Can't just use the public IPs, since we're dependent on DO's load balancers
- Droplets with custom images can't use IPv6

I ran `ifconfig` on the Linux host that I span up for testing, it looks like the private IP address is on an `eth1` network interface
<img width="556" alt="Screen Shot 2021-04-22 at 12 39 27 PM" src="https://user-images.githubusercontent.com/14321983/115752909-906e3500-a368-11eb-96ec-8879b4555aa9.png">

Here's the unikernel's output:
<img width="438" alt="Screen Shot 2021-04-22 at 12 46 08 PM" src="https://user-images.githubusercontent.com/14321983/115753039-b693d500-a368-11eb-811b-e766f521bc66.png">

I know responding to this might not be high on your priority list, but I'm excited to start using these if there's a solution. Thanks very much!

Edit: Modifying `hi.js` to return the results of Node's `os.networkInterfaces()` gives this:
<img width="279" alt="Screen Shot 2021-04-22 at 5 18 13 PM" src="https://user-images.githubusercontent.com/14321983/115786889-65e3a280-a38f-11eb-9094-d5430388693f.png">

francescolavra:
[...]

PR #1370, opened 2022-09-03
Added some configurable options for ProxMox API via CloudConfig

asyslinux:
[1356](https://github.com/nanovms/ops/issues/1356#issuecomment-1236103004) - Had some questions

1. Added full support of configurable options (CloudConfig) with backward compatibility:

Arch (Not work correctly through ProxMox API, but may be need later for other Clouds) (default: "" ; autodetect by ProxMox)
Machine (Not used if empty) (default: "" ; autodetect by ProxMox)
Sockets (default: "1")
Cores (default: "1")
Numa (default: "0")
Memory (default: "512M")
BridgeName (BridgeName0) (default: "vmbr0")
BridgeName1 (default: "") (Not used yet)
IsoStorageName (default: "local")
StorageName (default: "local-lvm")
Onboot (default: "0")
Protection (default: "0")

2. Added support for timestamps in instance names (if not used -i argument from command line).
3. Also fixed and extended some checks.
4. Added import dep of "github.com/docker/go-units" in proxmox/proxmox_instance.go for converting human-readable M,G memory settings to bytes for ProxMox

Short example of configuration file:
```
{
    "Dirs": ["static"],
    "NameServers": ["172.21.1.1"],
    "CloudConfig": {
        "BridgeName": "vmbr0",
        "ImageName": "mytest",
        "IsoStorageName": "local",
        "StorageName": "local-lvm"
    },
    "RunConfig": {
        "Accel": true,
        "InstanceName": "mytest",
        "ImageName": "mytest",
        "IPAddress": "172.21.1.75",
        "NetMask": "255.255.255.0",
        "Gateway": "172.21.1.1",
        "Verbose": true,
        "ShowErrors": true,
        "ShowWarnings": true
    }
}
```
Full example of configuration file:
```
{
    "Dirs": ["static"],
    "NameServers": ["172.21.1.1"],
    "CloudConfig": {
        "Arch": "x86_64",
        "Machine": "q35",
        "Sockets": 2,
        "Cores": 4,
        "Numa": true,
        "Memory": "4096M",
        "BridgeName0": "vmbr0",
        "ImageName": "mytest",
        "IsoStorageName": "local",
        "StorageName": "local-lvm",
        "Protection": false,
        "Onboot": true
    },
    "RunConfig": {
        "Accel": true,
        "InstanceName": "mytest",
        "ImageName": "mytest",
        "IPAddress": "172.21.1.75",
        "NetMask": "255.255.255.0",
        "Gateway": "172.21.1.1",
        "Verbose": true,
        "ShowErrors": true,
        "ShowWarnings": true
    }
}
```

Now with fully customizable base options, you can already normally use ops with ProxMox. It may be necessary to add some more options to the command line arguments, but this is after agreement with the Author of ops utility.

### DIFF
--- a/proxmox/proxmox_instance.go
+++ b/proxmox/proxmox_instance.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"strconv"
 
+	"github.com/docker/go-units"
 	"github.com/nanovms/ops/lepton"
 	"github.com/olekukonko/tablewriter"
 )
@@ -43,7 +44,7 @@ func (p *ProxMox) getNextID() string {
 		fmt.Println(err)
 	}
 
-	err = p.CheckResultType(body, "getnextid")
+	err = p.CheckResultType(body, "getnextid", "")
 	if err != nil {
 		return ""
 	}
@@ -56,16 +57,124 @@ func (p *ProxMox) getNextID() string {
 
 // CreateInstance - Creates instance on Proxmox.
 func (p *ProxMox) CreateInstance(ctx *lepton.Context) error {
+
+	var err error
+
 	config := ctx.Config()
 
 	nextid := p.getNextID()
 
+	instanceName := config.RunConfig.InstanceName
+
+	archType := config.CloudConfig.Arch
+	machineType := config.CloudConfig.Machine
+	socketsNum := config.CloudConfig.Sockets
+	coresNum := config.CloudConfig.Cores
+	numaStr := strconv.FormatBool(config.CloudConfig.Numa)
+	memoryHmn := config.CloudConfig.Memory
 	imageName := config.CloudConfig.ImageName
+	storageName := config.CloudConfig.StorageName
+	isoStorageName := config.CloudConfig.IsoStorageName
+	bridgeName := config.CloudConfig.BridgeName
+	bridgeName0 := config.CloudConfig.BridgeName0
+	onbootStr := strconv.FormatBool(config.CloudConfig.Onboot)
+	protectionStr := strconv.FormatBool(config.CloudConfig.Protection)
+
+	// Check CloudConfig options
+
+	if archType == "" {
+		archType = "x86_64"
+	}
+
+	if socketsNum == 0 {
+		socketsNum = 1
+	}
+
+	socketsStr := strconv.FormatInt(int64(socketsNum), 10)
+
+	if coresNum == 0 {
+		coresNum = 1
+	}
+
+	coresStr := strconv.FormatInt(int64(coresNum), 10)
+
+	if numaStr == "true" {
+		numaStr = "1"
+	} else {
+		numaStr = "0"
+	}
+
+	if memoryHmn == "" {
+		memoryHmn = "512M"
+	}
+
+	memoryInt, err := units.RAMInBytes(memoryHmn)
+	if err != nil {
+		return err
+	}
+
+	memoryInt = memoryInt / 1024 / 1024
+
+	memoryStr := strconv.FormatInt(memoryInt, 10)
+
+	if storageName == "" {
+		storageName = "local-lvm"
+	}
+
+	if isoStorageName == "" {
+		isoStorageName = "local"
+	}
+
+	if bridgeName0 == "" {
+		if bridgeName != "" {
+			bridgeName0 = bridgeName
+		} else {
+			bridgeName0 = "vmbr0"
+		}
+	}
+
+	err = p.CheckStorage(storageName, "images")
+	if err != nil {
+		return err
+	}
+
+	err = p.CheckStorage(isoStorageName, "iso")
+	if err != nil {
+		return err
+	}
+
+	err = p.CheckBridge(bridgeName0)
+	if err != nil {
+		return err
+	}
+
+	if onbootStr == "true" {
+		onbootStr = "1"
+	} else {
+		onbootStr = "0"
+	}
+
+	if protectionStr == "true" {
+		protectionStr = "1"
+	} else {
+		protectionStr = "0"
+	}
 
 	data := url.Values{}
 	data.Set("vmid", nextid)
-	data.Set("name", imageName)
-	data.Set("net0", "model=virtio,bridge=vmbr0")
+	data.Set("name", instanceName)
+	// Not work correctly through ProxMox API (Uses auto detecting by ProxMox)
+	// data.Set("arch", archType)
+	if machineType != "" {
+		data.Set("machine", machineType)
+	}
+	data.Set("sockets", socketsStr)
+	data.Set("cores", coresStr)
+	data.Set("numa", numaStr)
+	data.Set("memory", memoryStr)
+	data.Set("net0", "model=virtio,bridge="+bridgeName0)
+	data.Set("onboot", onbootStr)
+	data.Set("protection", protectionStr)
 
 	req, err := http.NewRequest("POST", p.apiURL+"/api2/json/nodes/"+p.nodeNAME+"/qemu", bytes.NewBufferString(data.Encode()))
 	if err != nil {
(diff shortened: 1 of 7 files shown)


eyberg:
so we now have 15 different infrastructure targets, i think it'd be best if cloudconfig/runconfig remain as generic as possible (eg: a given var should be more or less equal across target platforms), open to suggestion on this but kinda wondering if it might make sense to introduce optional platform specific config

we could provide a new interface that supplies custom config pertinent to each platform

for example all for these (which don't have equivalents in other platforms):

```
BridgeName (BridgeName0) (default: "vmbr0")
IsoStorageName (default: "local")
StorageName (default: "local-lvm")
Onboot (default: "0")
Protection (default: "0")
```

could all be a part of a proxmox specific config so we don't pollute all the other platform's config

proxmox/proxmox_instance.go
```diff
@@ -56,16 +57,124 @@ func (p *ProxMox) getNextID() string {
...
+		numaStr = "1"
+	} else {
+		numaStr = "0"
+	}
+
+	if memoryHmn == "" {
+		memoryHmn = "512M"
+	}
+
+	memoryInt, err := units.RAMInBytes(memoryHmn)
```
eyberg, line 111: we're hoping to use less imports (and could use a deepclean on the ones that exist) but if possible it be good to just copy in whatever method you need here and plop it in here https://github.com/nanovms/ops/blob/master/lepton/helpers.go

proxmox/proxmox_instance.go
```diff
@@ -56,16 +57,124 @@ func (p *ProxMox) getNextID() string {
...
+	machineType := config.CloudConfig.Machine
+	socketsNum := config.CloudConfig.Sockets
+	coresNum := config.CloudConfig.Cores
+	numaStr := strconv.FormatBool(config.CloudConfig.Numa)
+	memoryHmn := config.CloudConfig.Memory
 	imageName := config.CloudConfig.ImageName
+	storageName := config.CloudConfig.StorageName
+	isoStorageName := config.CloudConfig.IsoStorageName
+	bridgeName := config.CloudConfig.BridgeName
+	bridgeName0 := config.CloudConfig.BridgeName0
```
eyberg, line 79: what's the rationale for 2 of these? the expectation for multiple nics?

proxmox/proxmox_instance.go
asyslinux: Ok, i`ll be move part of code (units.RAMInBytes) to lepton/helpers.go, and will remove import.

proxmox/proxmox_instance.go
asyslinux: Yes, because often for production apps need to be have 2 nics, later i want add bridgeName1 for secondary nic. And not only use with DHCP, also will need add a static settings IPs too for secondary nic.

asyslinux:
> so we now have 15 different infrastructure targets, i think it'd be best if cloudconfig/runconfig remain as generic as possible (eg: a given var should be more or less equal across target platforms), open to suggestion on this but kinda wondering if it might make sense to introduce optional platform specific config
> 
> we could provide a new interface that supplies custom config pertinent to each platform
> 
> for example all for these (which don't have equivalents in other platforms):

Are you mean like this? Ok, i will try to add separate ProxmoxConfig part, please wait a little, but this is broke backward compatibility:

```
{
    "Dirs": ["static"],
    "NameServers": ["172.21.1.1"],
    "CloudConfig": {
        "Arch": "x86_64",
        "Machine": "q35",
        "Sockets": 2,
        "Cores": 4,
        "Numa": true,
        "Memory": "4096M",
    },
    "ProxmoxConfig": {
        "BridgeName0": "vmbr0",
        "ImageName": "mytest",
        "IsoStorageName": "local",
        "StorageName": "local-lvm",
        "Protection": false,
        "Onboot": true
    },
    "RunConfig": {
        "Accel": true,
        "InstanceName": "mytest",
        "ImageName": "mytest",
        "IPAddress": "172.21.1.75",
        "NetMask": "255.255.255.0",
        "Gateway": "172.21.1.1",
        "Verbose": true,
        "ShowErrors": true,
        "ShowWarnings": true
    }
}
```

asyslinux:
I changed the code as you indicated. Now configuration files is like as:
Full example of configuration file:
```
{
    "Dirs": ["static"],
    "NameServers": ["172.21.1.1"],
    "CloudConfig": {
        "ImageName": "mytest",
        "Arch": "x86_64",
        "Machine": "q35",
        "Sockets": 2,
        "Cores": 4,
        "Numa": true,
        "Memory": "4096M"
    },
    "ProxmoxConfig": {
        "BridgeName0": "vmbr0",
        "IsoStorageName": "local",
        "StorageName": "local-lvm",
        "Protection": false,
        "Onboot": true
    },
    "RunConfig": {
        "Accel": true,
        "InstanceName": "mytest",
        "ImageName": "mytest",
        "IPAddress": "172.21.1.75",
        "NetMask": "255.255.255.0",
        "Gateway": "172.21.1.1",
        "Verbose": true,
        "ShowErrors": true,
        "ShowWarnings": true
    }
}
```
Short example of configuration file:
```
{
    "Dirs": ["static"],
    "NameServers": ["172.21.1.1"],
    "CloudConfig": {
        "ImageName": "mytest"
    },
    "ProxmoxConfig": {
        "BridgeName": "vmbr0",
        "IsoStorageName": "local",
        "StorageName": "local-lvm"
    },
    "RunConfig": {
        "Accel": true,
        "InstanceName": "mytest",
        "ImageName": "mytest",
        "IPAddress": "172.21.1.75",
        "NetMask": "255.255.255.0",
        "Gateway": "172.21.1.1",
        "Verbose": true,
        "ShowErrors": true,
        "ShowWarnings": true
    }
}
```

types/config.go
```diff
@@ -146,6 +155,15 @@ type ProviderConfig struct {
 	// you can use to pass role information to an EC2 instance when the instance starts.
 	InstanceProfile string
 
+	// Machine specifies the type of machine (pc or q35) (Used only for ProxMox yet)
+	Machine string `cloud:"machine"`
```
francescolavra, line 159: The ProviderConfig struct already has a "Flavor" member, which is used to specify a machine type for other cloud providers (such as for example aws and gcp). It seems to me you could use that instead of adding a new field to the struct.

asyslinux: > The ProviderConfig struct already has a "Flavor" member, which is used to specify a machine type for other cloud providers (such as for example aws and gcp). It seems to me you could use that instead of adding a new field to the struct.

Maybe for Proxmox, in general, transfer all fields to ProxmoxConfig, and not use CloudConfig at all?

         "ImageName": "mytest",
         "Arch": "x86_64",
         "Machine": "q35",
         "Sockets: 2
         "Cores: 4
         "Numa": true
         "Memory": "4096M"

It just seems to me that if you add settings in the course of development, then you can add a bunch of additional settings to any providers. And in ProxmoxConfig, I have displayed far from everything that is configured there via Api.

asyslinux (commented):
Latest commit revert changes in CloudConfig struct, and all new fields is created in ProxmoxConfig struct. This is necessary to simplify the config for end users of ProxMox, and not to interfere with extra fields in the CloudConfig structure.

proxmox/proxmox_instance.go
eyberg: so we apparently do support multiple nics, https://github.com/nanovms/nanos/pull/1461 , https://github.com/nanovms/ops/issues/1009 however, ops needs more support to enable this fully, in particular network settings for static assignment need to be added in ops but this should prob be done in a different pr

types/config.go
```diff
@@ -80,6 +80,9 @@ type Config struct {
 	// attach/detach.
 	ProgramPath string
 
+	// ProxmoxConfig configures various attributes about the ProxMox provider.
+	ProxmoxConfig ProxmoxConfig
```
eyberg, line 84: So this is the right path, however, since we have 15 different infrastructure targets I don't think we want to include 15 separate ones into each top level config.

I was thinking this could be implemented via an interface so each target infrastructure can have its own cloud specific config but we only have one reference to it from here.

asyslinux: I'm afraid that a lot of code rework will be required in many places with the config, because everything is initially taken from c.Config, if I understand correctly. I'm not sure this is necessary, because provider selected from command line (-t arg) and if has correct error checking in providers, then user can not use wrong config. But this is my opinion, I just wanted to help get the ops working with proxmox.

eyberg: the problem is the same as the original one - this is all config for a specific target platform that can be arbitrarily large and it doesn't make sense to include for other users who won't use those platforms, i can prob take a look at this in the nxt few days as I do see your need but I just wouldn't want someone else to come in and start adding in the same thing for gcp, aws, azure, etc.

we have the notion of a Provider interface here -> https://github.com/nanovms/ops/blob/master/lepton/provider.go#L21 - I think adding the config into this so each provider can have one would be a good way to go about implementing this

asyslinux: No problem. It's up to you, I can wait if you rewrite it the way you want it yourself. You will make what architecture you need yourself for processing configs. When you implement, just reject the pull request and take what is required from my repository, that I wrote for proxmox and even that you can change to your liking.

eyberg: so as an example would this work for you? https://github.com/nanovms/ops/compare/master...targetConfig - that way that config is re-usable by any target platform and can be arb. large, in the past when we've had a number of target specific vars (such as Azure) we've utilized env vars which would also be an option here

asyslinux: Yes, good solution. Big Thanks.

eyberg:
chatted offline - closing in favor of https://github.com/nanovms/ops/pull/1373

PR #1378, opened 2022-09-10
Always enable AWS EnaSupport, and use t3.micro AWS instance flavor by default

paulreimer:
If I understand correctly, current nanos works with `EnaSupport=true` on both Nitro-based and non-Nitro-based instance types. Therefore, this PR removes the Nitro-instance detection mechanism and enables EnaSupport unconditionally.


paulreimer:
I am new to `nanos` / `ops`; this might be a breaking change in some ways that I have not anticipated, and it is also not perfectly backwards compatible with previous nanos versions (for example, prior to https://github.com/nanovms/nanos/pull/1383).

eyberg:
the reason we look at instance types is that different instances have different requirements (network adapters (in this case ena), different architectures, etc.) - we do need to know what type of instance someone is deploying to in many cases,

as for ena itself we unfortunately can't turn that on for every instance type un-conditionally: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking-ena.html#ena-requirements - what we can do is if a certain instance type supports it we could enable it by default

PR #1390, opened 2022-10-18
[:tada:] add env var interpolation to config files

zeroecco:
This feature PR allows the JSON configuration file to interpolate environmental variables allowing templating for a multitude of reasons including but not limited to:
* staging versus production deployments
* secrets stored in the JSON file

### DIFF
--- a/cmd/flags_config.go
+++ b/cmd/flags_config.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
 	"strings"
 
@@ -52,8 +53,8 @@ func unWarpConfig(file string, c *types.Config) (err error) {
 	if err != nil {
 		log.Fatalf("error reading config: %v\n", err)
 	}
-
-	return ConvertJSONToConfig(data, c)
+	loadedEnvJSON := os.ExpandEnv(string(data))
+	return ConvertJSONToConfig([]byte(loadedEnvJSON), c)
 }
 
 // ConvertJSONToConfig converts a byte array to an object of type configuration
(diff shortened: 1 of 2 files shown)


eyberg:
I'm generally in favor of this and something we already wanting to add as I attached a tkt to it - I also did a quick search of all the pkgs i have locally and nothing turned up '$' surprisingly

i know software that would use this (commonly seen in java applications espc) would already do the interpolation when ran in the shell which we wouldn't support regardless

I'm just wondering if there are examples where $ or ${} might appear in a config or package.manifest and we wouldn't want it to turn into empty string

barring any concerns from @nanovms/engineering perhaps adding a little env bool to turn this behavior on would eliminate that concern?

eg if os.Getenv("interop") != "" {...} else {...}

zeroecco:
I can add this flag

eyberg:
I think if you add that flag then we could go ahead and merge this; not sure if it's needed but it would prevent any accidental changes

cmd/flags_config.go
```diff
@@ -51,10 +51,14 @@ func (flags *ConfigCommandFlags) MergeToConfig(c *types.Config) (err error) {
 func unWarpConfig(file string, c *types.Config) (err error) {
 	data, err := ioutil.ReadFile(file)
 	if err != nil {
-		log.Fatalf("error reading config: %v\n", err)
+		log.Fatalf("error reading config: %v\nIf you are trying to use interoplation in your\nconfig's JSON set 'ops_render_config=true' in your ENV", err)
+	}
+	if os.Getenv("ops_render_config") == "true" {
```
eyberg, line 56: you could do a != "" to take any value but this is fine

eyberg (approved):

PR #1395, opened 2022-12-13
show labels attached to images

eyberg:
not sure if we should show these off/on by default but should definitely be available


francescolavra (commented):
Looks good to me

PR #1401, opened 2023-01-03
pkg(from-docker): reset/overwrite docker image Entrypoint

rinor:


eyberg:
hi @rinor - can you provide some background on what you're wanting to do here?

rinor:
If the docker image has configured an Entrypoint command, our ops command Cmd will be appended to that and we will end up executing

Entrypoint_command Cmd_ops_command that will be interpreted as args to entrypoin and starting the container may fail and the whole process fails.

Here I’m resetting the Entrypoint command (existing or not) so we start the container with only ops needed Cmd

eyberg (approved):

eyberg:
ok thanks!

Issue #1200, opened 2021-11-23
Adding ops JSON output to allow build web frontend


dzonerzy:
Would be nice if ops could output information like instances and so on in a JSON format this way would be possible to create a web frontend by wrapping the ops command.

eyberg:
this exists for many operations already like https://github.com/nanovms/ops/blob/master/gcp/gcp_instance.go#L210

dzonerzy:
I don't know if already exist for onprem images , but that would be awesome

eyberg:
proposal here would probably just be to tack on a '--json' flag to any of the commands vs default of table output

eyberg:
need to remove the RunConfig.JSON bool and put it somewhere else (possibly in a new struct that interacts with the various flags)

PR #1415, opened 2023-01-29
add global json flag support for listing do images and instances

geoah:
Hey, this is a continuation of https://github.com/nanovms/ops/pull/1240 that enabled the DO list instances and images commands to use the global `--json` flag.

I'm just playing around with ops, and needed json support for do commands, and it seemed easier to make a PR than make a ticket to ask for support for this in DO, hope that's ok.

~geoah

### DIFF
--- a/digitalocean/digital_ocean_image.go
+++ b/digitalocean/digital_ocean_image.go
@@ -2,6 +2,7 @@ package digitalocean
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


eyberg:
this is fine although I don't think the JSON bool belongs in runconfig as that really deals with a particular instance configuration not cli output that can span not just instances/images but pretty much anything; this ties into https://github.com/nanovms/ops/issues/1200 

we'd prob refactor this in the future to some output config of some kind

PR #1417, opened 2023-01-30
refactor: print error and exit instead of panics

Truenya:
panics replaced with printing error and Exit, mostly
In some cases, like network configuration, replaced with just message about error.


eyberg:
a few things:

1) I wouldn't default to colorize output - there are quite a few tools that consume output from ops - colorizing is really only for end-users

2) I wouldn't name that wrapper Fatal since that is being replaced and has a specific meaning in go stdlib

3) we already have a logger here, https://github.com/nanovms/ops/blob/master/log/logger.go so we shouldn't duplicate existing logic

lepton/package.go
```diff
@@ -373,19 +377,23 @@ func ExtractPackage(archive, dest string, config *types.Config) {
 		case tar.TypeDir:
 			if _, err := os.Stat(target); err != nil {
 				if err := os.MkdirAll(target, 0755); err != nil {
-					panic(err)
+					fmt.Printf("Failed to create directory %s", err)
```
francescolavra, line 380: It would be good to print the directory path as well, i.e. `target`

Truenya: done

qemu/qemu_unix.go
```diff
@@ -508,7 +508,7 @@ func generateMac() string {
 	octets := make([]byte, 6)
 	_, err := rand.Read(octets)
 	if err != nil {
-		panic(err)
+		fmt.Printf("Failed generate Mac, error is: %s", err)
```
eyberg, line 511: exit

gcp/gcp.go
```diff
@@ -13,7 +14,6 @@ import (
 	"github.com/nanovms/ops/types"
 	"golang.org/x/oauth2/google"
 	compute "google.golang.org/api/compute/v1"
-	"google.golang.org/api/dns/v1"
```
eyberg, line 16: i'm guessing you have a linter that removed this cause there was no reference to it;

it actually does have a reference though - the pkg name is 'dns' - you can add this line back and prefix it w/ 'dns' like the compute one above it

lepton/image.go
```diff
@@ -62,11 +62,11 @@ func addDNSConfig(m *fs.Manifest, c *types.Config) {
 	}
 	err := os.WriteFile(resolv, data, 0644)
 	if err != nil {
-		panic(err)
+		fmt.Printf("Failed save dns in temporary resolv.conf, error: %s", err)
```
eyberg, line 65: since this was a panic need to exit here

lepton/image.go
```diff
@@ -62,11 +62,11 @@ func addDNSConfig(m *fs.Manifest, c *types.Config) {
...
 	}
 	err = m.AddFile("/etc/resolv.conf", resolv)
 	if err != nil {
-		panic(err)
+		fmt.Printf("Failed add resolv.conf, error: %s", err)
 	}
```
eyberg, line 70: since this was a panic need to exit here

lepton/image.go
```diff
@@ -77,11 +77,11 @@ func addHostName(m *fs.Manifest, c *types.Config) {
 	data := []byte("uniboot")
 	err := os.WriteFile(hostname, data, 0644)
 	if err != nil {
-		panic(err)
+		fmt.Printf("Failed save hostname tmp file, error: %s", err)
```
eyberg, line 80: since this was a panic need to exit here

lepton/image.go
```diff
@@ -77,11 +77,11 @@ func addHostName(m *fs.Manifest, c *types.Config) {
...
 	}
 	err = m.AddFile("/proc/sys/kernel/hostname", hostname)
 	if err != nil {
-		panic(err)
+		fmt.Printf("Failed add hostname, error: %s", err)
 	}
```
eyberg, line 85: since this was a panic need to exit here

lepton/image.go
```diff
@@ -95,11 +95,11 @@ func addPasswd(m *fs.Manifest, c *types.Config) {
 	data := []byte("root:x:0:0:root:/root:/bin/nobash")
 	err := os.WriteFile(passwd, data, 0644)
 	if err != nil {
-		panic(err)
+		fmt.Printf("Failed save passwd in temporary file, error: %s", err)
```
eyberg, line 98: since this was a panic need to exit here

lepton/image.go
```diff
@@ -95,11 +95,11 @@ func addPasswd(m *fs.Manifest, c *types.Config) {
...
 	}
 	err = m.AddFile("/etc/passwd", passwd)
 	if err != nil {
-		panic(err)
+		fmt.Printf("Failed add passwd, error: %s", err)
 	}
```
eyberg, line 103: since this was a panic need to exit here

lepton/package.go
```diff
@@ -295,10 +295,11 @@ func GetPackageManifestFile() string {
 func PackageManifestChanged(fino os.FileInfo, remoteURL string) bool {
 	res, err := http.Head(remoteURL)
 	if err != nil {
-		if err, ok := err.(net.Error); ok {
+		var netError *net.Error
+		if errors.Is(err, *netError) {
 			fmt.Printf(constants.WarningColor, "missing internet?, using local manifest.\n")
 		} else {
-			panic(err)
+			fmt.Printf("probably bad URL: %s, got error %s", remoteURL, err)
```
eyberg, line 302: since this was a panic need to exit here

gcp/gcp.go
Truenya: done

lepton/image.go
Truenya: done

lepton/image.go
Truenya: done

lepton/image.go
Truenya: done

lepton/image.go
Truenya: done

lepton/image.go
Truenya: done

lepton/image.go
Truenya: done

lepton/package.go
Truenya: done

eyberg (approved):

PR #1424, opened 2023-02-05
chores(cleanup): remove folders [data, examples]

rinor:


eyberg:
I think it's a good idea to remove the binaries from the tree (forgot they were in here) and (def can remove the example stuff) 

although there is a test that is using that webg binary @ https://github.com/nanovms/ops/blob/master/lepton/ldd_linux_test.go#L13 and unfort it looks like it hasn't been firing since it has an invalid build tag; I think we could probably get rid of that in lieu of that ls test which is going to provide similar test coverage

eyberg:
thanks!

PR #1432, opened 2023-02-18
chore(lepton): remove lepton/homedir.go

rinor:


eyberg:
there was a reason why this was done this way but I'm having trouble finding the reference, https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/os/file.go;l=489 states android/ios being problematic but we don't really target those platforms so that's a non-issue - I think at least at one point it might've been because of getent? 

also, this is called a lot (see: https://github.com/nanovms/ops/issues/1095 ) so we'd like to keep the memoization regardless

rinor:
> also, this is called a lot (see: https://github.com/nanovms/ops/issues/1095 ) so we'd like to keep the memoization regardless

That's one of the "issues" I was aiming to solve in the long run, since while `cmd` and test packages can call/use that function directly, the other ones, e.g: provider packages, shouldn't need homedir directly (it can be already available/passed).

At the moment, imo, most of the stuff assumes that `ops` cmd/binary is the sole user of all/most of the packages, but in the future I would like to decouple some stuff so we could use the api/pkg cleanly (not that we can't use them right now) from other apps, e.g:  a clean api could be used with [dagger](https://github.com/dagger/dagger).

Closing this right now.

eyberg:
makes sense, this is exported so you should be able to call it from wherever, ops is already integrated into multiple other binaries && plugins fyi (home is used for storing lots of things both for running/building locally but also for storing releases, meta related to imgs/pkgs/etc.) and we would like the api to be fully composable as well (one of the reasons why we are starting to add protobufs)

as for dagger it should be fine for building images

PR #1434, opened 2023-02-18
chore(cleanup): remove constants/constants.go

rinor:
will get back to those strings and errors, but right now just cleaning stuff up.

### DIFF
--- a/cmd/error.go
+++ b/cmd/error.go
@@ -1,21 +1,19 @@
 package cmd
 
 import (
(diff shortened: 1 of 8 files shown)


eyberg:
so this is really a colors vs. no colors thing - I think that really should be a choice belonging to the end-user except for when the output is clearly being ingested via another program - I think it'd make the most sense to have a global/config level flag that users can turn on/off as desired

PR #1435, opened 2023-02-18
chore(cleanup): remove wsl/wsl.go

rinor:


eyberg:
I think these helpers probably need to stay since they can be accessed from multiple providers - what could be changed is isolating the code like we do with qemu where we break it up into platform specific files where go will only compile for the end platform target as shown in https://github.com/nanovms/ops/tree/master/qemu

PR #1436, opened 2023-02-18
chore(cleanup): testutils/mock_context.go

rinor:


eyberg:
it's a toss-up whether someone would use that helper again but in this case you yanked more lines out then put in so we'll take it

PR #1439, opened 2023-03-08
initial IBM cloud

eyberg:


provider/ibm/ibm_instance.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	table.Render()
+
+	return nil
+}
+
+// DeleteInstance deletes instance from IBM
+func (v *IBM) DeleteInstance(ctx *lepton.Context, instanceID string) error {
+	uri := "https://us-south.iaas.cloud.ibm.com/v1/instances/$instance_id?version=2023-02-28&generation=2"
```
francescolavra, line 218: Should use the region from the config instead of hardcoding us-south

provider/ibm/ibm_instance.go
```diff
@@ -0,0 +1,266 @@
...
+
+	"github.com/olekukonko/tablewriter"
+)
+
+func extractRegionFromZone(zone string) string {
+	s := strings.Split(zone, "-")
+	return s[0] + "-" + s[1]
+}
+
+// CreateInstance - Creates instance on Digital Ocean Platform
```
francescolavra, line 25: Digital Ocean -> IBM

PR #1445, opened 2023-03-25
build(provider): add build tags to disable providers

rinor:
the final implementation is like https://github.com/nanovms/ops/pull/1445#issuecomment-1493407963

---

By default nothing changes and all providers are compiled and available.

However now we have build tags to disable each provider (except `onprem`). i.e: the following command builds only `onprem` and all other providers are disabled.
```sh
go build -tags 
noaws,noazure,nodigitalocean,nogcp,nohyperv,noibm,nolinode,nooci,noopenshift,noopenstack,noproxmox,noupcloud,novbox,novsphere,novultr
```
---
available tags:
- no**aws**
- no**azure**
- no**digitalocean**
- no**gcp**
- no**hyperv**
- no**ibm**
- no**linode**
- no**oci**
- no**openshift**
- no**openstack**
- no**proxmox**
- no**upcloud**
- no**vbox**
- no**vsphere**
- no**vultr**

### DIFF
--- a/provider/aws/aws.go
+++ b/provider/aws/aws.go
@@ -1,3 +1,5 @@
+//go:build aws || !onlyprovider
+
 package aws
 
 import (
--- a/provider/aws/aws_dns.go
+++ b/provider/aws/aws_dns.go
@@ -1,14 +1,15 @@
+//go:build aws || !onlyprovider
+
 package aws
 
 import (
 	"strconv"
 	"time"
 
-	"github.com/nanovms/ops/types"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/route53"
 	"github.com/nanovms/ops/lepton"
+	"github.com/nanovms/ops/types"
 )
 
 // FindOrCreateZoneIDByName searches for a DNS zone with the name passed by argument and if it doesn't exist it creates one
--- a/provider/aws/aws_image.go
+++ b/provider/aws/aws_image.go
@@ -1,3 +1,5 @@
+//go:build aws || !onlyprovider
+
 package aws
 
 import (
@@ -24,7 +26,6 @@ import (
 	"github.com/nanovms/ops/log"
 	"github.com/nanovms/ops/types"
 	"github.com/olekukonko/tablewriter"
-
 	"github.com/schollz/progressbar/v3"
 )
 
--- a/provider/aws/aws_image_test.go
+++ b/provider/aws/aws_image_test.go
@@ -1,3 +1,5 @@
+//go:build aws || !onlyprovider
+
 package aws_test
 
 import (
--- a/provider/aws/aws_instance.go
+++ b/provider/aws/aws_instance.go
@@ -1,3 +1,5 @@
+//go:build aws || !onlyprovider
+
 package aws
 
 import (
--- a/provider/aws/aws_instance_group.go
+++ b/provider/aws/aws_instance_group.go
@@ -1,7 +1,10 @@
+//go:build aws || !onlyprovider
+
 package aws
 
 import (
 	"fmt"
+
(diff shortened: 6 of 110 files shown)


eyberg:
I suppose I'm not opposed to this although the ergonomics feel reversed - if you only wanted 'onprem' than seems to me like it should exclude everything else vs having to specify each one you don't want

also, it'd be nice to have something like this in nanos as that's where you'd actually get more potential benefits (less code to abuse and smaller size would help boot time)

rinor:
By default only `onprem` provider is compiled and available. It's not possible to disable this one.
i.e: the following command builds only `onprem` and all other providers are disabled.
```sh
go build
```
However now we have build tags to enable each provider (or all of them). 
i.e: the following command builds all available providers.
```sh
go build -tags allproviders
```
i.e: the following command builds `aws` provider (also `onprem` that is always enabled).
```sh
go build -tags aws
```

---
available tags:
- **allproviders** (build all available providers)
- specific provider(s)
  - **aws**
  - **azure**
  - **digitalocean** (or **do**)
  - **gcp**
  - **hyperv**
  - **ibm**
  - **linode**
  - **oci**
  - **openshift**
  - **openstack**
  - **proxmox**
  - **upcloud**
  - **vbox**
  - **vsphere**
  - **vultr**

rinor:
> also, it'd be nice to have something like this in nanos as that's where you'd actually get more potential benefits (less code to abuse and smaller size would help boot time)

Are you referring to things like the followings to be put behind build flags, or you meant smth else?
From a quick check the candidates for build flags may be:
```
https://github.com/nanovms/nanos/tree/master/src/aws
https://github.com/nanovms/nanos/tree/master/src/hyperv
https://github.com/nanovms/nanos/tree/master/src/vmware
https://github.com/nanovms/nanos/tree/master/src/xen
https://github.com/nanovms/nanos/tree/master/src/drivers/gve.*
```
and maybe
```
https://github.com/nanovms/nanos/tree/master/src/gdb
```

eyberg:
sorry I wasn't super clear, I feel that we want the default build to still include support for everything, if an individual user wanted to trim it down to one or two providers I think we'd be fine allowing code/tags to do that but I don't see a ton of gain in making that the default; as I said earlier doing this in nanos would make a lot more sense - most of ops end users are using ops to build and deploy but not necessarily run so most of the trimming benefits would come from doing this in nanos where un-used code can be removed from the run-time preventing abuse and allow shorter boot times, however, I think that would entail a lot more than build tags and would probably be worth thinking out a bit w/a proposal (possibly via discussions) vs jumping into code

rinor:
By default nothing changes and all providers are compiled and available.

However now we have build tags to select a subset of provider(s)  if required (except `onprem` that is built always). 

- the following command builds everything 
```sh
go build
```

- the following command builds only `onprem` and all other providers are disabled.
```sh
go build -tags onlyprovider
```

- the following command builds only `aws` and all other providers are disabled (except `onprem` that is built always).
```sh
go build -tags onlyprovider,aws
```

To select the providers to build you need to always use `onlyprovider` tag accompanied with provider tags:
```
go build -tags onlyprovider,aws,azure,do,gcp,ibm,linode,upcloud,vultr
```

---
available tags:
- **onlyprovider**
  - **aws**
  - **azure**
  - **digitalocean**
  - **gcp**
  - **hyperv**
  - **ibm**
  - **linode**
  - **oci**
  - **openshift**
  - **openstack**
  - **proxmox**
  - **upcloud**
  - **vbox**
  - **vsphere**
  - **vultr**

rinor:
I understand that this may not seem a useful feature, so please feel free to close this pr without merging it.
The main reason behind this work is/was related to ops `daemon` feature, ofc depending on how far that service goes/evolves.

PR #1458, opened 2023-05-10
Update aws_image.go

zeroecco:
note that c6i is nitro based per docs here:
https://aws.amazon.com/ec2/instance-types/c6i/


francescolavra (commented):
Looks good to me

PR #1462, opened 2023-05-15
add json options to vultr

unusualevent:
this copies the same layout as the Digitalocean provider. You previously discussed moving the JSON setting around.

Couldn't get this to build locally because the go mirror seems to be misbehaving around docker/distribution V2.8.0. Contemplated upgrading to v2.8.1 in the PR but maybe not.


eyberg:
I think this is fine for now - thanks!

PR #1487, opened 2023-06-22
config: remove VirtfsShares from user file config

rinor:
based on https://github.com/nanovms/ops-documentation/pull/411#issuecomment-1602193432


francescolavra (commented):
Looks good to me

PR #1518, opened 2023-08-25
AWS images: set ENA support to true regardless of instance type

francescolavra:
The EnaSupport flag of the ec2.RegisterImageInput structure indicates the the OS in an image has an ENA driver, not that it requires an ENA device; thus, an image flagged as ENA-capable can run just fine on AWS instances without an ENA adapter (such as non-Nitro instance types).
This change removes the NitroInstanceTypes map and the GetEnaSupportForFlavor() function, and sets the EnaSupport flag to true regardless of the instance type. This allows creating images that can run on newer Nitro instance types such as m7a, m7i and r7iz.


eyberg (approved):
LGTM, although i'm not sure if we can remove it or not according to https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#RegisterImageInput ```    // This option is supported only for HVM AMIs. Specifying this option with a
    // PV AMI can make instances launched from the AMI unreachable.
```

francescolavra:
AWS images registered by Ops are always HVM AMIs, regardless on the instance types - https://github.com/nanovms/ops/blob/4f8ac6fa66e4f75d76bb2cf8eb20c78f42a5ea81/provider/aws/aws_image.go#L153

eyberg:
ah ok, lgtm

PR #1521, opened 2023-09-13
gcp fw rules are at vpc not subnet level

eyberg:


francescolavra (commented):
LGTM. The same fix should also be applied to the two buildFirewallRule() calls in createInstanceTemplate() in gcp_instance_group.go

PR #1527, opened 2023-09-30
allow nanos-version to be used in cross-arch

eyberg:


radiosilence:
Why not just use -arm64 as local folder to make sure the arch is specifically 64 bit?

eyberg:
we do do this already, just not everything conforms to it yet,

the set of possible runs are:

* run x86 on arm64
* run arm64 on x86
* run x86 on x86
* run arm64 on arm64

so we want sane defaults everywhere (eg: default to user's native platform) but allow them to override as they see fit

radiosilence:
I get it but the PR basically adds `-arm` and I figured there would be no harm in making it more explicit :)

eyberg:
i see what you are saying - nanos is 64bit only - we don't have plans for ever adding 32bit; this pr is kinda dead regardless; i'll need to revisit it once i find some time

PR #1530, opened 2023-10-20
allow basevolumesz to be set on instance creation alone instead of im…

eyberg:
…g creation for gcp

### DIFF
--- a/provider/gcp/gcp_instance.go
+++ b/provider/gcp/gcp_instance.go
@@ -55,17 +55,30 @@ func (p *GCloud) CreateInstance(ctx *lepton.Context) error {
 	instanceName := c.RunConfig.InstanceName
 	canIPForward := c.RunConfig.CanIPForward
 
+	cimg := &compute.AttachedDiskInitializeParams{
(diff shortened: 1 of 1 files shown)


provider/gcp/gcp_instance.go
```diff
@@ -55,17 +55,30 @@ func (p *GCloud) CreateInstance(ctx *lepton.Context) error {
...
+	cimg := &compute.AttachedDiskInitializeParams{
+		SourceImage: imageName,
+	}
+
+	if c.BaseVolumeSz != "" {
+		size, err := lepton.GetSizeInGb(c.BaseVolumeSz)
+		if err != nil {
+			return fmt.Errorf("cannot get volume size: %v", err)
+		}
+		c.BaseVolumeSz = "" // create minimum-sized local volume
```
francescolavra, line 67: This is not needed (it's only needed when the command being executed creates a local image/volume file, which is not the case for `ops instance create`).

francescolavra (approved):
LGTM

PR #1542, opened 2023-11-14
gcp(tags): backward compatible extended usage

rinor:
I need a bit more control over what goes where.
The existing behavior is kept the same, unless one uses the new specific definitions.

- this config behaves the same as before:
```json
{
  "CloudConfig": {
    "Tags": [
      {
        "key": "instance-and-image",
        "value": "instance-image"
      }
    ]
  }
}
```
- new config options - all options (you can set to `true` only what you need)
```json
{
  "CloudConfig": {
    "Tags": [
      {
        "key": "everywhere",
        "value": "image-instance-metadata-valueAtnetwork",
        "image_label": true, // added to image labels upon image build
        "instance_label": true, // added to instance labels upon instance creation
        "instance_metadata": true // added to instance custom metadata upon instance creation
        "instance_network": true,  // tag VALUE is added to instance network tags
      }
    ]
  }
}
```

### DIFF
--- a/provider/gcp/gcp.go
+++ b/provider/gcp/gcp.go
@@ -35,14 +35,21 @@ type GCloudOperation struct {
 	operationType string
 }
 
-func buildGcpTags(tags []types.Tag) map[string]string {
-	labels := map[string]string{}
+func buildGcpLabels(tags []types.Tag, imageORinstance string) map[string]string {
+	labels := map[string]string{"createdby": "ops"}
 	for _, tag := range tags {
+		switch imageORinstance {
+		case "instance":
+			if !tag.IsInstanceLabel() {
+				continue
+			}
+		case "image":
+			if !tag.IsImageLabel() {
+				continue
+			}
+		}
 		labels[tag.Key] = tag.Value
 	}
-
-	labels["createdby"] = "ops"
-
 	return labels
 }
 
--- a/provider/gcp/gcp_image.go
+++ b/provider/gcp/gcp_image.go
@@ -103,7 +103,7 @@ func (p *GCloud) CreateImage(ctx *lepton.Context, imagePath string) error {
 
 	rb := &compute.Image{
 		Name:         c.CloudConfig.ImageName,
-		Labels:       buildGcpTags(ctx.Config().CloudConfig.Tags),
+		Labels:       buildGcpLabels(ctx.Config().CloudConfig.Tags, "image"),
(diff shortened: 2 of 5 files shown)


eyberg:
tags are heavily used by quite a few people/orgs so I do like the backwards compatibility of this; one change that i might suggest if you want to break it up into varying types - create a new tag struct and stick an enum in it to define the type with the default choice being the old style (eg: used for everything)

rinor:
> tags are heavily used by quite a few people/orgs so I do like the backwards compatibility of this; one change that i might suggest if you want to break it up into varying types - create a new tag struct and stick an enum in it to define the type with the default choice being the old style (eg: used for everything)

what if we add an **extended** `attribute` to tags, containing the new specific attributes, that if present and set, will activate and use new functionality, i.e:


1. **Existing behavior, remains unchanged**, if configured as always

```json
{
  "CloudConfig": {
    "Tags": [
      {
        "key": "instance-and-image",
        "value": "instance-image"
      }
    ]
  }
}
```

2. **Equivalent behavior/result** if using the new config

```json
{
  "CloudConfig": {
    "Tags": [
      {
        "key": "instance-and-image",
        "value": "instance-image",
        "attribute": {
          "image_label": true,
          "instance_label": true
        }
      }
    ]
  }
}
```

3. If preferred and desired, `feedback needed here`, we could even treat an **empty attribute list** as not set, so this config would have the **same behavior/result as (1. and 2.)**

```json
{
  "CloudConfig": {
    "Tags": [
      {
        "key": "instance-and-image",
        "value": "instance-image",
        "attribute": {}
      }
    ]
  }
}
```

4. And this would be **different from (3.)**, since as soon as you **add at least one field** inside `attribute` (`true` or `false` it doesn't matter), you opt-in for the new explicit config:
```json
{
  "CloudConfig": {
    "Tags": [
      {
        "key": "this-goes-nowhere",
        "value": "non-existent",
        "attribute": {
          "image_label": false,
          "instance_label": false,
          "instance_network": false,
          "instance_metadata": false
        }
      }
    ]
  }
}
```

In the meantime, I will update the code so we can check what it could look like

eyberg:
yeh - makes sense - i'm good w/these changes - is this ready to merge?

rinor:
> yeh - makes sense - i'm good w/these changes - is this ready to merge?

yes, will push also the updated ops docs a bit later

PR #1543, opened 2023-11-14
gcp(volumes): configurable autoattach upon instance creation

rinor:
- `RunConfig.AttachVolumeOnInstanceCreate` - will attempt to attach already created disk(s)/volume(s) to the instance that is being created with `ops instance create --config config.json ....`

```json
{
  "Mounts": {
    "volume_name_or_id": "/db"
  },
  "RunConfig": {
    "AttachVolumeOnInstanceCreate": true
  }
}
```

**Note**: this feature is available on **ops config file only**

### DIFF
--- a/provider/gcp/gcp_instance.go
+++ b/provider/gcp/gcp_instance.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


provider/gcp/gcp_instance.go
```diff
@@ -117,6 +118,61 @@ func (p *GCloud) CreateInstance(ctx *lepton.Context) error {
 		}
 	}
 
+	// Attach disk(s) upon instance creation if RunConfig.AttachVolumeOnInstanceCreate is set
+	for mntSrc, mntDst := range c.Mounts {
+		if !c.RunConfig.AttachVolumeOnInstanceCreate {
+			continue
+		}
```
francescolavra, line 125: This check can be done just once, if the attach option is false we don't need to execute the for loop.

provider/gcp/gcp_instance.go
```diff
@@ -117,6 +118,61 @@ func (p *GCloud) CreateInstance(ctx *lepton.Context) error {
...
+
+		// VirtFS VirtIO 9p mount - skip it
+		if strings.HasPrefix(mntSrc, "/") {
+			continue
```
francescolavra, line 129: 9P mounts are for onprem instances only, specifying a 9P mount directive for a GCP instance doesn't make sense, so I wouldn't do this check. If the user inserts a mntSrc with a leading '/' I guess it will result in a disk not found error

francescolavra (commented):
LGTM

Issue #591, opened 2020-08-25
local user-created pkgs seem to hate symlinks for some reason


eyberg:
we added support for symlinks for load/run on existing packages but user-created packages loaded via '-l' seem to have a problem

eyberg:
https://github.com/nanovms/ops/pull/1476

PR #1544, opened 2023-11-16
Minor pkg command changes

BunningsWarehouseOfficial:
Made a few changes to `pkg` to address some things I came across while learning to use it recently.

- `pkg from-run` and `pkg from-pkg` were supposed to require the `-v` version flag to append to the package name, but the wrong variable was checked
- changed `pkg from-docker` to not append the version using a different format to `from-run` and `from-pkg`
- Made `pkg from-docker` copy over the Program using the executable's actual name instead of just "program". Loading a package by default saves an image using the name of its executable. This resulted in the image being shown with the name “program” under `ops image list`, which as someone new to OPS was very confusing and led me astray regarding an unrelated problem I was having. It also means the images from different local packages overwrite each other. I didn’t know about the `--imagename` flag for `ops pkg load` at the time, but I think the default should be to use the original executable’s name as defined with e.g. the `--file` flag of `ops pkg from-docker`

Another change I wanted to make was to make `ops pkg from-docker` with the `--copy` flag copy over symlink files (the links themselves, not what they point to) into the package's sysroot. This was because the Python Docker image I was using tried and failed to use the version compatibility link `libz.so.1` (among others) despite the linked to file, `libz.so.1.2.11`, being present in the sysroot, resulting in a crash. I figured it’d be in the spirit of “copying the whole file system” (as --copy is described) for something like a Python package to just work off the bat. I thought the change would be fairly trivial, but was mistaken; couldn't get it to work in a reasonable amount of time. Keen to hear thoughts on this

### DIFF
--- a/cmd/cmd_pkg.go
+++ b/cmd/cmd_pkg.go
@@ -457,7 +457,7 @@ func cmdFromPackage(cmd *cobra.Command, args []string) {
 	}
 
 	version, _ := flags.GetString("version")
-	if newpkg == "" {
+	if version == "" {
 		fmt.Println("missing new version")
 		os.Exit(1)
 	}
@@ -485,7 +485,7 @@ func cmdFromRun(cmd *cobra.Command, args []string) {
 	}
 
 	version, _ := flags.GetString("version")
-	if newpkg == "" {
+	if version == "" {
 		fmt.Println("missing new version")
 		os.Exit(1)
 	}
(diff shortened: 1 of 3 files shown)


eyberg:
thanks!

for the symlink issue - feel free to open up a new issue - as i didn't see anything that was tied to this directly

i feel like local pkgs had that support added but wasn't sure cause it was marked wip - https://github.com/nanovms/ops/issues/591 && https://github.com/nanovms/ops/pull/1476

the from-docker code looks like it's missing symlink support - you can find some examples in fs/manifest.go if you grep for ModeSymlink - for fixing it - ideally we could re-use/consolidate code that might already exist

PR #1547, opened 2023-11-18
gcp(volumes): ops volume create-image/create-from (GCP only)

rinor:
---

- **ops volume create-image**
```
$ ops volume create-image --help
create volume image only

Usage:
  ops volume create-image <image_name> [flags]

Flags:
  -d, --data string   volume image data source
  -h, --help          help for create-image
  -s, --size string   volume image size when restored onto a persistent disk
```

---

- **ops volume create-from**

```
$ ops volume create-from --help 
create volume from existing image

Usage:
  ops volume create-from <image_name> <volume_name> [flags]

Flags:
  -h, --help          help for create-from
  -s, --size string   volume image size when restored onto a persistent disk
```

---

**Note**: Don't like the api very much, but at the same time didn't want to touch the existing commands. Needed this so drafting it here for feedback (or in case someone else needs it).

### DIFF
--- a/cmd/cmd_volume.go
+++ b/cmd/cmd_volume.go
@@ -17,7 +17,7 @@ func VolumeCommands() *cobra.Command {
 	cmdVolume := &cobra.Command{
 		Use:       "volume",
 		Short:     "manage nanos volumes",
-		ValidArgs: []string{"create, list, delete, attach"},
+		ValidArgs: []string{"create-image, create-from, create, list, delete, attach"},
 		Args:      cobra.OnlyValidArgs,
 	}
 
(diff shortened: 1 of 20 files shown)


eyberg:
not sure of the use-case here? can you elaborate? you want to be able to create an image from an existing volume and create a volume from an image?

rinor:
The idea is to create an volume image and later on create different instance disk volumes from that same image that is alread uploaded and available. I.e I'm testing/creating 100 instances that need a persistent volume and don't see a reason to locally create/upload the same thing 100 times.

eyberg:
ah ok, GCP has a disk clone feature that might be a better fit for that? https://cloud.google.com/compute/docs/disks/create-disk-from-source

rinor:
closing this until I find a better less intrusive format/api

PR #1553, opened 2023-11-28
gcp(volumes): ops volume create - add disk image ops (GCP only)

rinor:
a different, less intrusive, take on https://github.com/nanovms/ops/pull/1547
This allows additional flow for non-boot disk volumes on clouds (gcp only atm).

- `ops volume create <volume_name> --data ./data_dir --size 1G` - functionality remains unchanged. As a reminder this command performs the following main actions:
  **a.** create a disk image (format fs, include data, compress, upload, create disk image) - named `<volume_name>`
  **b.** use disk image created at (**a**) to create an instance attachable disk - named `<volume_name>`
  
New flow allows to execute those actions separately:
- `ops volume create <image_name> --data ./data_dir --size 1G --image-only` - `size`, `data` are still optional, `--image-only` is the new flag. Basically this performs only the previous step (**a**)
  - create a disk image (format fs, include data, compress, upload, create disk image) - named `<image_name>`
  
- `ops volume create <volume_name> --size 2G --image <image_name>` - `data` is not allowed here, `size` is still optional,  `--image` is the new flag. Basically this performs only the previous step (**b**)
  - use an **existing** disk image, named ` <image_name>`, created previously with `--image-only` flag or from the original command,  to create an instance attachable disk named `<volume_name>`

---

Use cases:
- Create an empty TFS formated disk image, and use that image as source for any/all instance(s) disk volumes that expect/want additional empty volumes, unrelated to a specific application.
- The same applies to images with data, useful when you have a lot of data included inside the volume, you "upload" (create the disk image) just once and use that to "update"/"create" the instance volumes.


eyberg:
for cloning large volumes is it not possible to use the link I provided earlier? https://cloud.google.com/compute/docs/disks/create-disk-from-source#go

rinor:
> for cloning large volumes is it not possible to use the link I provided earlier? https://cloud.google.com/compute/docs/disks/create-disk-from-source#go

it is, and I can add the extra flow to use as source an existing disk (not image), but here I'm not "cloning" anything and using "images" as sources. And that link provides the additional sources disk, snapshot.

eyberg:
I guess i'm still a bit unclear as to why would you want to use an image as the source vs using a volume as the source since that is what you want as the result? isn't it the use-case that you want to copy a large volume on GCP's side w/out having to upload it each time?

PR #1564, opened 2023-12-09
feat(onprem): volume ls / volume cp - added

rinor:
- `ops volume ls`
```
list files and directories in volume

Usage:
  ops volume ls <volume_name:volume_uuid> <path> [flags]

Flags:
  -l, --long-format   use a long listing format
```
- `ops volume cp`
```
copy files from volume to local filesystem

Usage:
  ops volume cp <volume_name:volume_uuid> <src>... <dest> [flags]

Flags:
  -L, --dereference   always follow symbolic links in volume
  -r, --recursive     copy directories recursively
```

### DIFF
--- a/cmd/cmd_volume.go
+++ b/cmd/cmd_volume.go
@@ -2,13 +2,13 @@ package cmd
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"path"
 	"strconv"
 	"strings"
 
 	"github.com/nanovms/ops/fs"
+	"github.com/nanovms/ops/log"
 	"github.com/nanovms/ops/provider/onprem"
 	"github.com/nanovms/ops/types"
 
@@ -21,7 +21,7 @@ func VolumeCommands() *cobra.Command {
 	cmdVolume := &cobra.Command{
 		Use:       "volume",
 		Short:     "manage nanos volumes",
-		ValidArgs: []string{"create, list, delete, attach"},
+		ValidArgs: []string{"create, list, delete, attach, tree, ls, cp"},
 		Args:      cobra.OnlyValidArgs,
 	}
 
@@ -38,6 +38,8 @@ func VolumeCommands() *cobra.Command {
 	cmdVolume.AddCommand(volumeAttachCommand())
 	cmdVolume.AddCommand(volumeDetachCommand())
 	cmdVolume.AddCommand(volumeTreeCommand())
+	cmdVolume.AddCommand(volumeLsCommand())
+	cmdVolume.AddCommand(volumeCopyCommand())
 	return cmdVolume
 }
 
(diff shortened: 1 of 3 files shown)


cmd/cmd_volume.go
```diff
@@ -225,13 +227,130 @@ func volumeTreeCommandHandler(cmd *cobra.Command, args []string) {
 	reader.Close()
 }
 
+func volumeLsCommand() *cobra.Command {
+	var cmdLs = &cobra.Command{
+		Use:   "ls <volume_name:volume_uuid> <path>",
```
francescolavra, line 232: `path` should be enclosed in square brackets, since it's an optional argument

cmd/cmd_volume.go
```diff
@@ -225,13 +227,130 @@ func volumeTreeCommandHandler(cmd *cobra.Command, args []string) {
...
+		Short: "list files and directories in volume",
+		Run:   volumeLsCommandHandler,
+		Args:  cobra.MinimumNArgs(1),
+	}
+	flags := cmdLs.PersistentFlags()
+	flags.BoolP("long-format", "l", false, "use a long listing format")
+	return cmdLs
+}
+
+func volumeLsCommandHandler(cmd *cobra.Command, args []string) {
```
francescolavra, line 242: This function is identical to imageLsCommandHandler() except for the call to getLocalVolumeReader() instead of getLocalImageReader(). I would make a more generic `imageLs(flags *pflag.FlagSet, args []string, reader *fs.Reader)` function and use it in both imageLsCommandHandler() and volumeLsCommandHandler().

cmd/cmd_volume.go
```diff
@@ -225,13 +227,130 @@ func volumeTreeCommandHandler(cmd *cobra.Command, args []string) {
...
+		Run:   volumeCopyCommandHandler,
+		Args:  cobra.MinimumNArgs(3),
+	}
+	flags := cmdCopy.PersistentFlags()
+	flags.BoolP("recursive", "r", false, "copy directories recursively")
+	flags.BoolP("dereference", "L", false, "always follow symbolic links in volume")
+	return cmdCopy
+}
+
+func volumeCopyCommandHandler(cmd *cobra.Command, args []string) {
```
francescolavra, line 302: Likewise, I would make a more generic `imageCp(flags *pflag.FlagSet, args []string, reader *fs.Reader)` function and use it in both imageCopyCommandHandler() and volumeCopyCommandHandler().

cmd/cmd_volume.go
rinor: the plan was to refactor those a bit late for reuse, but will do it now, thanks

cmd/cmd_volume.go
rinor: that wouldn't work normally, because of the usage of this stuff that I would delete completely. However patched just enough to make it work for what's needed. https://github.com/nanovms/ops/blob/4ea40fb7dac1b01606a8032a81699ffee3420bf6/cmd/cmd_root.go#L66-L111

francescolavra (commented):
LGTM

PR #1577, opened 2024-02-13
for onprem sort virt-fs mounts so attachment_ids line correspond to labels across image/instance creation

eyberg:


provider/onprem/onprem_volume.go
```diff
@@ -242,9 +243,15 @@ func AddVirtfsShares(config *types.Config) error {
 	query := make(map[string]string)
 	virtfsShares := make(map[string]string)
 
+	ocp := make(map[string]string)
```
francescolavra, line 246: This ocp map is never used

qemu/qemu_unix.go
```diff
@@ -510,9 +511,26 @@ func (q *qemu) setConfig(rconfig *types.RunConfig) error {
 	q.addOption("-device", "virtio-rng-pci")
 
 	disks := 1
-	for _, hostDir := range rconfig.VirtfsShares {
-		q.addOption("-virtfs", fmt.Sprintf("local,path=%s,mount_tag=%d,security_model=none,multidevs=remap,id=hd%d", hostDir, disks, disks))
-		disks++
+
+	if len(rconfig.VirtfsShares) > 0 {
+
+		// if using virtual fs shares sort first.
```
francescolavra, line 517: We don't actaully need to sort the virtfs shares. As long as there is a fixed correspondence between virtual IDs and host dirs (which there is, due to the sorting done in `AddVirtfsShares`), we can simply iterate over rconfig.VirtfsShares, as in:
```
for diskId, hostDir := range rconfig.VirtfsShares {
		q.addOption("-virtfs", fmt.Sprintf("local,path=%s,mount_tag=%s,security_model=none,multidevs=remap,id=hd%s", hostDir, diskId, diskId))
}
```
The order in which these `-virtfs` command line options are added to the qemu command lone may vary, but that doesn't matter.

Issue #1578, opened 2024-02-19
Instance tags through config.json in DO


hnord-vdx:
I have added tags in `CloudConfig` (e.g. "function": "webserver") but they dont show on digital ocean. Is this not possible or am i doing something wrong?

francescolavra:
Ops currently does not assign tags in `CloudConfig` to Digital Ocean instances, but this feature can definitely be added in the future.

hnord-vdx:
Alright, thanks! This feature would be quite useful, especially when combined with DO load balancers, as it would seamlessly direct the load to the new instance.

PR #1580, opened 2024-02-19
Added CloudConfig tags to DO instance

hnord-vdx:
This allows for tags set in `CloudConfig` to also be set on DO instance (see https://github.com/nanovms/ops/issues/1578). I implemented it in a way where the value of tag is used if key is empty. This allows for tags without `:` when desired

### DIFF
--- a/provider/digitalocean/digital_ocean_instance.go
+++ b/provider/digitalocean/digital_ocean_instance.go
@@ -51,6 +51,17 @@ func (do *DigitalOcean) CreateInstance(ctx *lepton.Context) error {
 		})
 	}
 
+	tags := make([]string, 0, len(config.CloudConfig.Tags)+2)
+	tags = append(tags, opsTag, imageName)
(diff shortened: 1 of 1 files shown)


francescolavra (commented):
Looks good to me

provider/digitalocean/digital_ocean_instance.go
```diff
@@ -51,14 +51,25 @@ func (do *DigitalOcean) CreateInstance(ctx *lepton.Context) error {
 		})
 	}
 
+	tags := make([]string, 0, len(config.CloudConfig.Tags)+2)
```
eyberg, line 54: what's this +2 for?

you can just do a ```tags:= []string{}```

hnord-vdx: +2 is for the two default tags added (see line 55). Your option definitely looks nicer, I can amend my pull request if you want. I just wanted to exactly allocate the required memory!

PR #1582, opened 2024-02-26
Use VPC field for DO as VPCUUID

hnord-vdx:
Adding the `VPCUUID` field allows you to set a VPC of the droplet in the config.json. Ideally you could reference it by name instead of UUID but thats a DO limitation. At least this way you can set VPC at instance creation without having to do a snapshot and then adding it to non-default VPC

Note: We would need to mention this in the documentation as its not obvious


eyberg:
I believe you can get a list of vpc name:uuid via the vpcs list call as shown here: https://github.com/digitalocean/godo/blob/main/vpcs_test.go#L65 and then you can pass in the name - this is how we do it for other providers as well, is it possible to add that?

PR #1584, opened 2024-02-26
Set VPC on instance creation using VPC name

hnord-vdx:
Adding the `VPCUUID` field allows you to set a VPC of the droplet in the config.json. This way you can set VPC at instance creation without having to do a snapshot and then adding it to non-default VPC

### DIFF
--- a/provider/digitalocean/digital_ocean_instance.go
+++ b/provider/digitalocean/digital_ocean_instance.go
@@ -61,11 +61,20 @@ func (do *DigitalOcean) CreateInstance(ctx *lepton.Context) error {
 		}
 	}
 
+	vpc, err := do.GetVPC(ctx, config.CloudConfig.VPC)
+	if err != nil {
+		return err
+	}
+	vpcUUID := ""
+	if vpc != nil {
+		vpcUUID = vpc.ID
+	}
+
 	createReq := &godo.DropletCreateRequest{
 		Name:    instanceName,
 		Size:    flavor,
 		Region:  config.CloudConfig.Zone,
-		VPCUUID: config.CloudConfig.VPC,
+		VPCUUID: vpcUUID,
 		Image: godo.DropletCreateImage{
 			ID: imageID,
 		},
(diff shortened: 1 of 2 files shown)


hnord-vdx:
@eyberg I see you are importing the `google/uuid` package perhaps we could add a check (`uuid.Parse(vpcName)`) if the name is a UUID and then skip the whole `GetVPC` round trip?  

Also, currently my implementation matches the LAST vpc based on the name. Maybe we want to match first VPC and warn using `Debugf` if there are other matches?

eyberg:
i'd assume the vpc name is unique and so you really only need the first match? also i'm not crazy about passing in the id optionally but it appears we already do that in aws https://github.com/nanovms/ops/blob/master/provider/aws/aws_network.go#L163 so i'm ok w/allowing that if that's easier

hnord-vdx:
Yeah, you are right. I just double checked vpc names have to be unique within a account so the zone check is not needed.

provider/digitalocean/digital_ocean_instance.go
```diff
@@ -61,10 +62,23 @@ func (do *DigitalOcean) CreateInstance(ctx *lepton.Context) error {
 		}
 	}
 
+	// if users set VPC uuid instead of name it avoids lookup
+	vpcUUID := config.CloudConfig.VPC
+	if _, err := uuid.Parse(vpcUUID); err != nil {
```
eyberg, line 67: can skip al of this if user didn't set it

provider/digitalocean/digital_ocean_network.go
```diff
@@ -0,0 +1,54 @@
...
+func (do *DigitalOcean) GetVPC(ctx *lepton.Context, vpcName string) (*godo.VPC, error) {
+
+	if vpcName == "" {
+		return nil, nil
+	}
+
+	page := 1
+	var vpc *godo.VPC
+
+	// ctx.Logger().Debug("getting all vpcs")
```
eyberg, line 23: yank

eyberg:
thanks!

Issue #1594, opened 2024-03-15
Cannot seem to push locally built package


radiosilence:
To make it simpler here's a Makefile that should in theory compile/build out my package, add it to ops, and then push it:

```make
PKGNAME=nano-web
PKGVERSION=0.1.0
PKGRELEASE=$(PKGNAME)_$(PKGVERSION)
RELEASEDIR=./release
PKGDIR=$(RELEASEDIR)/$(PKGRELEASE)

pkg-clean:
	rm -rf $(RELEASEDIR)

pkg-build:
	 GOOS=linux go build -o $(PKGDIR)/$(PKGNAME) main.go

pkg-create: pkg-clean
	mkdir -p $(PKGDIR)/sysroot
	mkdir -p $(PKGDIR)/sysroot/public
	./scripts/make-manifest.sh > $(PKGDIR)/package.manifest
	cp README.md $(PKGDIR)

pkg-add: pkg-create pkg-build
	ops pkg add $(PKGDIR) --name $(PKGRELEASE)

pkg-bundle: pkg-add
	cd $(RELEASEDIR); tar czvf $(PKGRELEASE).tar.gz $(PKGRELEASE)
	@echo "Release created: $(PKGDIR).tar.gz"

pkg-push: pkg-add
	ops pkg push $(PKGRELEASE)

pkg-load: pkg-add
	ops pkg load -l $(PKGRELEASE) -p 80

```

When I run `make pkg-push` I get this output:

```
make pkg-push   
rm -rf ./release
mkdir -p ./release/nano-web_0.1.1/sysroot
mkdir -p ./release/nano-web_0.1.1/sysroot/public
./scripts/make-manifest.sh > ./release/nano-web_0.1.1/package.manifest
cp README.md ./release/nano-web_0.1.1
GOOS=linux go build -o ./release/nano-web_0.1.1/nano-web main.go
ops pkg add ./release/nano-web_0.1.1 --name nano-web_0.1.1
nano-web_0.1.1
ops pkg push nano-web_0.1.1
no local package with the name nano-web_0.1.1 found
```

Have I missed a step somehow? It is slightly unclear if I am doing something wrong with regards to the docs.

If I use `ops pkg load -l` it works as expected:

```
 ops pkg load -l nano-web_0.0.1
running local instance
booting /Users/jc/.ops/images/nano-web ...
en1: assigned 10.0.2.15
⇨ public directory not found in: /
```

[...]

PR #1596, opened 2024-03-16
Fix local package regexp for packages with hyphen in name

radiosilence:
Currently there is a bug where because the regexp is liberal enough to allow either _ or - to delimit the break between name and version, a package with dash will fail.

So for instance, a package like

my-package_1.2.3

Will have a name of "my" and a version of "package_1.2.3".

Not wanting to change this behaviour by simply changing the regex to only accept _ as a delimeter, I have changed the regex to split at the *last* match.

Fixes: https://github.com/nanovms/ops/issues/1594

### DIFF
--- a/lepton/package.go
+++ b/lepton/package.go
@@ -27,7 +27,7 @@ import (
 // PackageSysRootFolderName is the name of package root folder
 const PackageSysRootFolderName = "sysroot"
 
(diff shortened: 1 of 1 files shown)


radiosilence:
Fixes https://github.com/nanovms/ops/issues/1594

eyberg (approved):

Issue #1604, opened 2024-04-09
[Mac M1] ops pkg list (question / bug)


galuszkak:
Hey,

I tried to play with ops but it's unclear to me if it works correctly.

So I did look on https://repo.ops.city/ and search for all `python` packages. 

I did the same in shell (I've added API Key using `ops pkg login`) but when I do:
```bash
➜  ~ ops pkg list --search python

+-----------+-------------+---------+----------+---------+--------------------------------+
| NAMESPACE | PACKAGENAME | VERSION | LANGUAGE | RUNTIME |          DESCRIPTION           |
+-----------+-------------+---------+----------+---------+--------------------------------+
| eyberg    | python      | 3.10.6  | c        | 3.10.6  | The Python programming         |
|           |             |         |          |         | language.                      |
+-----------+-------------+---------+----------+---------+--------------------------------+
```

So it's only 1 package not 9 how I saw it on https://repo.ops.city. Am I'm doing something wrong here?
<img width="629" alt="Screenshot 2024-04-09 at 12 54 29" src="https://github.com/nanovms/ops/assets/759579/34d4be73-b508-4c58-9f88-0d729181b8ae">

galuszkak:
OK I realized is because of the CPU architecture. Closing.

eyberg:
yes - x86 pkgs vastly outnumber arm ones only because those are the ones that most commonly get deployed to prod envs; if you see something that you'd like to see on arm though feel free to open an issue and we can possibly take a look

also, you can always set the arch flag via to run cross-arch too:

```
ops pkg load eyberg/sabledb:0.0.1 -p 6379 --arch=amd64
```

PR #1610, opened 2024-04-11
feat: add CPU Architecture for Package Search & List

galuszkak:
Hey,

thanks for great project!
Based on issues that I had in #1604 , I thought it would be good add CPU Arch column so it's more obvious what are those packages. Also there are scenarios that you could have both images locally in different architectures and ability to differentiate them is important.
I also thought that behavior in list should be changed to show all arch by default, and to only filter them if someone adds `--arch`.

Let me know if this PR is acceptable and if would be acceptable to change pkg list behaviour.

### DIFF
--- a/cmd/cmd_pkg.go
+++ b/cmd/cmd_pkg.go
@@ -234,6 +234,7 @@ func cmdListPackages(cmd *cobra.Command, args []string) {
 		row = append(row, pkg.Version)
 		row = append(row, pkg.Language)
 		row = append(row, pkg.Runtime)
+		row = append(row, pkg.Arch)
 		row = append(row, pkg.Description)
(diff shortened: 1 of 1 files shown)


eyberg:
pkg list doesn't have an arch flag yet but pkg search does

I think we had defaulted to the user's real arch w/the option of showing others so as not to confuse anyone; there are other clients that call into ops but as long as the correct arch is coming through that should be fine

> Also there are scenarios that you could have both images locally in different architectures and ability to differentiate them is important.

ops actually doesn't have support for packages w/same namespace/name/version but diff arch today (the repo does) - i have some work in a diff branch that i'm going to push soon to enable that

regardless - I don't see anything wrong w/this pr so will go ahead and merge

PR #1612, opened 2024-04-13
Find binaries with PATH

mayl:
Removes absolute paths, instead finds binaries using PATH resolution

### DIFF
--- a/cmd/cmd_network.go
+++ b/cmd/cmd_network.go
@@ -252,7 +252,7 @@ func killBridge(bridgeName string) {
 }
 
 func execCmd(cmdStr string) (output string, err error) {
-	ecmd := exec.Command("/bin/bash", "-c", cmdStr)
+	ecmd := exec.Command("bash", "-c", cmdStr)
 	out, err := ecmd.CombinedOutput()
 	if err != nil {
 		return
--- a/lepton/package.go
+++ b/lepton/package.go
@@ -570,7 +570,7 @@ func addToPackage(newconfig *types.Config, newpath string) {
 }
 
 func execCmd(str ...string) {
-	cmd := exec.Command("/bin/bash", "-c", strings.Join(str, " "))
+	cmd := exec.Command("bash", "-c", strings.Join(str, " "))
 	out, err := cmd.CombinedOutput()
 	if err != nil {
 		fmt.Println(err)
--- a/network/iproute_network_service.go
+++ b/network/iproute_network_service.go
@@ -12,7 +12,7 @@ type IprouteNetworkService struct {
 }
 
 func execCmd(cmdStr string) (output string, err error) {
-	cmd := exec.Command("/bin/bash", "-c", cmdStr)
+	cmd := exec.Command("bash", "-c", cmdStr)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
 		return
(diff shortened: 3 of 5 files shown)


mayl:
This builds and mostly works, but some of the commands don't work and I'm not sure if it's a `nixos` thing or an `ops` thing.

For example:

```
❯ ops network create
exit status 1
```

but running
```
❯ ops -j --show-debug --show-warnings --show-errors network create
exit status 1
```
doesn't give me more information to debug what's exiting with 1.

This is the same for me before and after this patch.

eyberg:
sorry, just coming back to this,

would just be using '/bin/sh' work here and we can remove the bash refs? none of these cmds afaik actually use anything bash specific, the longer term fix is to get rid of as much of the shelling out as possible

as for the error you are seeing - the network creation stuff either needs the user in a group w/privileges or suid or the like - it's for users that want to run their own orchestration; it could be more helpful w/the error msg though

mayl:
> sorry, just coming back to this,

Np! I'm working on this on hobby time so nights and weekends so don't expect super responsiveness from me either... 

> would just be using '/bin/sh' work here and we can remove the bash refs? none of these cmds afaik actually use anything bash specific, the longer term fix is to get rid of as much of the shelling out as possible

Sure, I was just carrying forward the bash/sh usage as it existed before, but we can make this all just `sh`.  

Is there a reason to run all this in a shell besides pipe-ing between commands?  Linking stdout to stdin looks like it could also be [setup in go](https://pkg.go.dev/os/exec@go1.22.2#Cmd.StdinPipe) and then there would be no dependency on any shell.

> 
> as for the error you are seeing - the network creation stuff either needs the user in a group w/privileges or suid or the like - it's for users that want to run their own orchestration; it could be more helpful w/the error msg though

Hmm, on my machine even running ops as root creates the same result for me.  I really don't _think_ it's related to these changes, I think it's more likely something in my `nix` machine's networking is setup to be immutable.  I can't really expect ops to handle that weirdness, I just don't want to inadvertently introduce something which breaks any of your other users...

eyberg:
any piping you see in those commands would have been done for connivence to make something work quickly - most of this stuff has equivalents in go that could/should be used instead; it's not just piping - it'd be best if any exec call were to be avoided if possible - there are a few of these that vastly easier to fix then the others (eg: mkdir/cp should be straight-forward; replacing the call to something like qemu not so much) - but i wouldn't expect his pr to do any of this - I think your original goal was just to have less dependence on the abs. path for bash;

you should know that at least for the cmd_network stuff you were looking at - that currently has deps on brctl && dnsmasq; another place uses arp && ifconfig

so it's not just bash that has some hard-coded deps

mayl:
> you should know that at least for the cmd_network stuff you were looking at - that currently has deps on brctl && dnsmasq; another place uses arp && ifconfig
> 
> so it's not just bash that has some hard-coded deps

Yes, but that's fine because it's resolved through `$PATH`, which `nix` has facilities to populate appropriately. I actually add these dependencies in the `nix` package as well. The problem from `nix`'s perspective is if you changed these to `/bin/brctl` for instance.  In that case it's not just sufficient for there to be a `brctrl` executable available, it also has to be literally located at `/bin/brctl`.

I noticed your most recent commit.  That looks like it could completely subsume this PR if you just swap `/bin/bash` for `bash` or `sh` in those two locations in tools.

mayl:
superseded by https://github.com/nanovms/ops/pull/1615

PR #1618, opened 2024-04-25
use --arch for ops pkg push cmd (with sanitisation for amd64)

radiosilence:
Basically this completes the loop for pushing packages with multiple arches via CLI.

Seeing as the current `--arch` flag accepts either `amd64` (as opposed to `x86_64`) or `arm64`, I'm manually changing this in the request data when it's pushed to the repo. 

Also falls back to `x86_64` if it's blank/unset

Also it slightly changes the behaviour when searching local packages, as previously it was erroring out if one of the existing arches didn't exist yet, so we just log it and continue instead.

### DIFF
--- a/cmd/cmd_pkg.go
+++ b/cmd/cmd_pkg.go
@@ -593,8 +593,11 @@ func pushCommandHandler(cmd *cobra.Command, args []string) {
 	}
 
 	private, _ := flags.GetBool("private")
+	arch, _ := flags.GetString("arch")
 	ns, name, version := api.GetNSPkgnameAndVersion(pkgIdentifier)
+
 	pkgList, err := api.GetLocalPackageList()
+
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


cmd/cmd_pkg.go
```diff
@@ -623,7 +624,7 @@ func pushCommandHandler(cmd *cobra.Command, args []string) {
 	}
 	defer os.RemoveAll(archiveName)
 
-	req, err := api.BuildRequestForArchiveUpload(ns, name, foundPkg, archiveName, private)
+	req, err := api.BuildRequestForArchiveUpload(ns, name, foundPkg, archiveName, private, arch)
```
eyberg, line 627: can you change this to pkgFlags.Parch() ?

then I believe you can get rid of the arch flag above as well

radiosilence: Oh shoot I just saw this, is the pr still relevant? I can sort it out 😃

eyberg: yeh - i'd just like to have everything use that one Parch() method, and then you can drop that other arch flag

radiosilence: Ok I'll do that when I'm by my laptop

PR #1628, opened 2024-06-11
update: move from direct http to linodego client for api requests

vkhamsao:
Description:
Updates:
- Adds linodego pkg for calls to linode api
- Adds stop, reboot, and delete functionality for instances

Fixes:
- Updated root pass to address Linode failing to provision due to password complexity requirements.

Mostly keeps the same logic in place just using linodego.  I can submit any doc updates if this is merged.

Thanks.


eyberg:
in general we'd prefer to have less dependencies than more - but you're not the first person that mentioned this w/ linode so will go ahead and merge

PR #1630, opened 2024-06-16
change default mac runner from x86 to arm64

eyberg:


eyberg:
circleci has phased out their x86 mac runners so i yanked the mac build from https://github.com/nanovms/ops/pull/1633 until it can support arm here

PR #1669, opened 2024-12-05
allow overriding default azure gallery name

eyberg:
since gallery names are globally unique you need a diff. one for other regions.


eyberg:
https://github.com/nanovms/ops-documentation/commit/f6a88ae834a211450facbe2fdf3b220ebc0c5561

PR #1678, opened 2025-01-14
converting aws provider to use aws-sdk-go-v2

chickencoding123:
I needed `AWS_ENDPOINT_URL` environment variable and apparently [aws-sdk-go](https://github.com/aws/aws-sdk-go) doesn't respect it. This v1 sdk also doesn't respect many other AWS environment variables. V1 sdk is in maintenance mode now. A simple fix on my part would've been to just pass this new value into `Session`, but I wanted to dig into go for the very first time, not even a hello world in the past :smile: 

I understand you may want to consider accepting this since at some point you'll have to upgrade the sdk anyways. If you do consider this contribution, then please run the internal tests to make sure nothing breaks.

Purpose of this PR is to upgrade the go sdk version to v2.


eyberg:
thanks! quite a bit of go for first time - i just tried this out for basic operations and seems good - i'll take a look more in a bit

eyberg (approved):

PR #1682, opened 2025-01-23
fix(provider_disabled): fix methods to include filtering capabilities…

rinor:
… and new InstanceStats method

- Add filter argument to ListImages method
- Add filter argument to GetImages method
- Introduce InstanceStats method for instance monitoring
- Update CreateVolume method to accept CloudVolume type


eyberg (approved):

eyberg:
thanks!

PR #1685, opened 2025-02-10
chore: network - use 'ip' instead of 'brctl' and 'ifconfig'

rinor:
- Replace `brctl` commands with `ip` commands for bridge management
- Update error messages for clarity on command failures
- Use `ip link show` to check device existence
- Create bridge with `ip link add` command
- Set bridge device IP address using `ip addr add`
- Bring bridge device up with `ip link set dev up`
- Bring bridge device down using `ip link set dev down`
- Delete bridge device using `ip link delete` command
- Check empty bridge with `ip link show master` command

### DIFF
--- a/cmd/cmd_network.go
+++ b/cmd/cmd_network.go
@@ -76,21 +76,31 @@ func createBridgedNetwork(bn string, subnet string) {
 		network = subnet
 	}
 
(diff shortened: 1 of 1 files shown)


eyberg (approved):

eyberg:
thanks!

PR #1688, opened 2025-02-18
feat: support image (ls, tree, cp) over bootfs

rinor:
- `ops image ls   --bootfs ...` - list over boot filesystem
- `ops image tree --bootfs ...` - tree over boot filesystem
- `ops image cp   --bootfs ...` - copy from boot filesystem
- Implement `NewReaderBootFS` for boot filesystem (`kernel` + `klib`) handling


eyberg:
useful - thanks!

PR #1691, opened 2025-02-28
chore: lepton.CreateArchive can rename files

### LINKED ISSUE #1149: someone reported getting duplicate disk files when using 'image create'
not sure where the 'disk.raw' is coming from or if it's a tmp file not being removed?

```
ops image create -c ~/config.json --package mysql_5.7.29 -i mysql-unikernel-test -t gcp
```

```
ls -ltr ~/.ops/images/
total 319448
-rw-r--r-- 2 user group 209715200 Aug 19 15:58 mysql-unikernel-test.img
-rw-r--r-- 2 user group 209715200 Aug 19 15:58 disk.raw
```

rinor:
resolve #1149

### DIFF
--- a/provider/azure/azure_image.go
+++ b/provider/azure/azure_image.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


provider/azure/azure_image.go
```diff
@@ -65,27 +64,6 @@ func (a *Azure) getArchiveName(ctx *lepton.Context) string {
...
-
-	err := os.Link(imagePath, symlink)
-	if err != nil {
-		return "", err
-	}
-
-	archPath := filepath.Join(filepath.Dir(imagePath), a.getArchiveName(ctx))
-	files := []string{symlink}
-
-	err = lepton.CreateArchive(archPath, files)
```
eyberg, line 84: was all of this accidentally removed?

rinor: I don't see the archive used anywhere in azure. That function returns the original raw image path.

PR #1705, opened 2025-06-08
adding in default options for inception

eyberg:


eyberg:
the address assignment should still be cleaned up i'm going to go ahead and merge this so it's not left hanging around

PR #1719, opened 2025-10-19
feat(provider,hetzner): create hetzner provider

0xE282B0:
This PR adds the basic functionality needed to deploy Nanos Unikernels on Hetzner Cloud.

### DIFF
--- a/provider/hetzner/hetzner_image.go
+++ b/provider/hetzner/hetzner_image.go
@@ -64,11 +64,11 @@ func (h *Hetzner) CustomizeImage(ctx *lepton.Context) (string, error) {
 	return imagePath, nil
 }
 
+// BuildImage creates a Hetzner-compatible image from the active configuration.
 func (h *Hetzner) BuildImage(ctx *lepton.Context) (string, error) {
 	c := ctx.Config()
-	if !c.Uefi {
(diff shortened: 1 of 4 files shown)


eyberg:
thanks! i just tested this out - i think the reason we never added it before was cause of the extra buildserver step but that didn't seem to take too long so I'd be happy to merge this

can you add the option to https://github.com/nanovms/ops/blob/master/cmd/flags_provider.go#L70 ?

I think the linter is just complaining about lack of exported function comments (i can take care of this if you want)

eyberg (approved):

provider/hetzner/hetzner_image.go
```diff
@@ -0,0 +1,521 @@
...
+
+// CustomizeImage returns image path with adaptations needed by cloud provider
+func (h *Hetzner) CustomizeImage(ctx *lepton.Context) (string, error) {
+	imagePath := ctx.Config().RunConfig.ImageName
+	return imagePath, nil
+}
+
+func (h *Hetzner) BuildImage(ctx *lepton.Context) (string, error) {
+	c := ctx.Config()
+	if !c.Uefi {
```
eyberg: if this is true for all their instance types I'd prob. just auto-enable this

PR #1781, opened 2026-05-29
refactor: replace interface{} with any (Go 1.18+)

MD-Mushfiqur123:
Replace `interface{}` with the `any` alias (available since Go 1.18) across the codebase.

Only type literal usages were changed (not interface definitions).


eyberg:
thnx

PR #1793, opened 2026-09-15
pkg: build a package from a dockerfile

AngeloRubens:
ops pkg from-dockerfile ./Dockerfile builds the Dockerfile, takes the file system and the
configuration out of the image it produces, and leaves a package behind that ops pkg load runs.

    ops pkg from-dockerfile ./Dockerfile --name myapp
    ops pkg load -l myapp -p 8080

What the image says of itself becomes the manifest: ENTRYPOINT and CMD become the program and its
arguments, ENV the environment, WORKDIR the working directory, EXPOSE the ports, and the size of
the file system the volume to hold it.

When the image starts a script
------------------------------

Most published images do. A unikernel is one process and cannot run a shell, so there is nothing
to be had from a docker-entrypoint.sh by reading it. The --resolve-entrypoint flag runs the image
under docker instead and watches what it actually starts:

    ops pkg from-dockerfile ./Dockerfile --resolve-entrypoint

The program nearest the top of the process tree is the one taken, with the argv and the
environment it was given -- a launcher that exports variables before handing over is common, and
the exports live only in the environment of what it started. Shells, tini, dumb-init, catatonit
and the s6 tools are seen through rather than taken. What the run leaves behind is kept, so an
image that sets itself up on first boot has that work done in the package rather than still
ahead of it.

Where nothing can be read -- nginx rewrites its own command line -- it says so and builds
nothing, rather than making a package that would not start.

The images it was built against
-------------------------------

Each is a real image, from the repository that publishes it, left alone. Every row is a machine
that booted and was asked for something.

| image | flags | what comes out |
|---|---|---|
| prometheus | | answers /-/healthy |
| etcd | | serves /version, keeps its write ahead log |
| influxdb | | answers on 8086 |
| caddy | | serves its pages |
| dotnet, aspnet | | answers on 8080 |
| spring boot | | answers on 8080 |
| helidon | | answers /greet |
| python, uvicorn | | answers on 8000 |
| tomcat | --resolve-entrypoint | answers on 8080 |
| wildfly | --resolve-entrypoint | answers on 8080 |
| grafana | --resolve-entrypoint | answers /login |
| redis | --resolve-entrypoint | answers on 6379 |
| node | --resolve-entrypoint | answers on 3000 |
| temurin, java | --resolve-entrypoint | answers on 8080 |
| tautulli | --resolve-entrypoint | puts its web server on 8181 |
| wordpress | --resolve-entrypoint | apache up and serving |
| nextcloud | --resolve-entrypoint | apache up and serving |
| memcached | --resolve-entrypoint | serves on 11211, on a kernel that reads uid from the manifest, https://github.com/nanovms/nanos/pull/2189 |

And nine shapes of Dockerfile, each a program that starts and says its piece: a static binary on
scratch, a dynamic one on debian, an alpine root file system, a distroless base, a multi stage
build driven by --target and --build-arg, the syntax directive BuildKit reads, an ENTRYPOINT in
shell form, one whose command is built out of variables, and a python image.

Twenty seven in all. Every one but memcached is a scenario in test/dockerfile, built and booted
on each change:

    OPS=$PWD/ops ./test/dockerfile/run.sh tomcat-resolved

Alongside them the suite keeps the refusals: an image whose program cannot be reached has to be
turned away with the reason, and that is asserted too.

Four things this found on the way
---------------------------------

Each is its own commit, and each stands without the rest of this:

- a package file system's top level links were extracted as copies of what they pointed at
- a package carrying a file the image left unreadable could not be extracted at all
- a container file system was copied without the modes and hard links the container had, and over
  /dev, /proc and /sys, which the kernel asserts on
- what a package says about the manifest was not carried into the image built from it

What it asks for
----------------

The build goes through the docker command when it is on the path, so that the BuildKit features
-- heredocs, RUN --mount, COPY --chmod, BUILDPLATFORM -- work as they do everywhere else, and
falls back to the daemon's own builder otherwise. The --resolve-entrypoint flag needs a docker
daemon, being a run of the image. Nothing else is new.

| flag | |
|---|---|
| dockerfile | the file to build, ./Dockerfile by default |
| --context | the build context, the directory of the Dockerfile by default |
| --build-arg | name=value, or name to take it from the environment |
| --target | the stage to build in a multi stage Dockerfile |
| --name, --version | the name and version of the package |
| --resolve-entrypoint | run the image to see what it starts |
| --resolve-timeout | seconds to watch it for, 60 by default |
| --keep-image | keep the docker image the build produces |


AngeloRubens:
Withdrawing this for now. As it stands the package carries the whole root file system of the docker image, operating system and all, and that is not what belongs in a unikernel image. Reworking it so that only what the program actually reaches goes in.

